PR #7, opened 2024-06-24
It is better to use the repository field in Cargo.toml

szabgab:
To allow [Crates.io](https://crates.io/) , [lib.rs](https://lib.rs/) and the [Rust Digger](https://rust-digger.code-maven.com/) to link to it. See [the manifest](https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field) for the explanation.


Valerioageno:
Thanks for spotting it. Could you please update it also on `tuono_lib` and `tuono_lib_macro` crates?

szabgab:
done

Valerioageno (approved):
Thanks!

PR #92, opened 2024-11-12
fix: development script watch loop

Valerioageno:


marcalexiei:
The loop is now resolved. The `dev` task ends and remain in watch.

However there is another issue: when editing a file inside the `apps/documentation` folder the watch process doesn't recompile the frontend part. Everything work if I restart the `dev` task but I'm assuming this is not expected.

BTW thanks for the quick support!

Valerioageno:
That's weird. I wouldn't expect it to reload, since the documentation doesn't use the workspace library but the deployed one (`"tuono": "0.12.0"`). It should act exactly like a normal `tuono` app, so it should reload only when watched with `tuono dev`.

PR #93, opened 2024-11-13
refactor: use `React.JSX` rather than global `JSX`

### LINKED ISSUE #84: React `JSX` global namespace is deprecated
Greetings,
I just tried to setup my first tuono project using
```shell
tuono new [NAME]
```

Everything runs correctly but I want to report that the documentation (e.g.: [Components page](https://tuono.dev/documentation/tutorial/components)) and template files (e.g.: `src/routes/index.tsx`) are using `JSX` global namespace from `@types/react` which is now deprecated:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/80e20dcf70052f5bacb051f641ea507e705aba5b/types/react/index.d.ts#L4302-L4306

```ts
declare global {
    /**
     * @deprecated Use `React.JSX` instead of the global `JSX` namespace.
     */
    namespace JSX {
```

---

If you wish I can setup a PR that import `React` in all the relevant template and documentation files.

marcalexiei:
## Context & Description

Close #84.

I remove all the usage of global `JSX` I was able to find and replaced them with the one included inside `React`.

> [!NOTE]
> I used typed named imports for it unless there there was already a `import * as React from 'react'`, 
> in that case I used `React.JSX` instead of adding another import.

### Additional related changes 

- I reaplaced few empty fragments (`<></>`) with `null`
- I enabled [`@typescript-eslint/no-deprecated`](https://typescript-eslint.io/rules/no-deprecated/) rule which can report when using `@deprecated` items

---

While working on this I got a few ideas about some improvements to repository maintenance and DX.
I'll open separate issue to discuss them later this week.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -49,7 +49,7 @@
     "@typescript-eslint/naming-convention": [
       "error",
       {
-		"selector": "typeParameter",
+        "selector": "typeParameter",
         "format": ["PascalCase"],
         "leadingUnderscore": "forbid",
         "trailingUnderscore": "forbid",
@@ -59,6 +59,7 @@
         },
       },
     ],
+    "@typescript-eslint/no-deprecated": "error",
     "@typescript-eslint/no-empty-function": "error",
     "@typescript-eslint/no-empty-interface": "error",
     "@typescript-eslint/no-explicit-any": "off",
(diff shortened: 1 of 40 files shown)


Valerioageno:
This is a great improvement. Thanks!

Valerioageno (approved):

PR #96, opened 2024-11-14
chore: add `apps/documentation` to `pnpm` workspaces

### LINKED ISSUE #95: Repo Maintenance: adding `apps/documentation` to `pnpm` workspaces
`apps/documentation` right now is ignored by root workspace setup so it requires a "standalone" pnpm install.   

I think that we can add this folder to the workspaces by:
- using `npm:tuono@{{version}}` notation to explicit that in these folder `tuono` must be downloaded from `npm` registry
- change the package name to avoid clash with the workspaces inside `packages` folder
   ```diff
   + "name": "tuono-documentation",
   - "name": "tuono",
   ```

---

### Pros

1. avoid duplication of some dev configuration / devDependencies from the root package (e.g.: ESLint) 
2. simplify project setup procedure (no need to write about separate `pnpm install` in different directories).
3. user can run check on local machine from the root without having to navigate across folders via file explorer or CLI.
    e.g.: I created #93 and format checks failed on `apps/documentation` folder on [CI run](https://github.com/Valerioageno/tuono/actions/runs/11811551359)

marcalexiei:
## Context & Description

Close #95.

`apps/documentation` is now part of `pnpm` workspaces.
All relevant scripts have been added to the root `package.json`.

> [!NOTE]
> Later I'm going to improve format also on `json` and `yml` files

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -15,52 +15,52 @@ jobs:
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
     timeout-minutes: 15
     runs-on: ubuntu-latest
-    
-    steps:    
-    - name: Check out code
-      uses: actions/checkout@v4
-      with:
-        fetch-depth: 2
 
-    - name: Setup Node.js environment
-      uses: actions/setup-node@v4
-      with:
-        node-version: 20
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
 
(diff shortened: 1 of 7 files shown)


.github/workflows/typescript.yml
```diff
@@ -15,52 +15,52 @@ jobs:
...
 
-    - name: Setup Node.js environment
-      uses: actions/setup-node@v4
-      with:
-        node-version: 20
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
```
Valerioageno, line 23: Cool

PR #100, opened 2024-11-15
ci: create `install-node-dependencies` action

### LINKED ISSUE #97: Repo maintenance: reusable action to setup NodeJS dependencies
Related to #96.

When working with the GitHub Actions workflows I noticed some issues:

## Setup `pnpm` via `npm i -g`

Some workflows setup `pnpm` using `npm i -g pnpm` rather than `pnpm/action-setup`.
I think we can always use `pnpm/action-setup` in conjunction with `setup-node` cache set to `pnpm` which might speed dependencies installation a little bit.

## `pnpm-lock.yaml` is not versioned

I noticed that `pnpm` lock file is ignored from versioning. 
IMHO this is a problem because someone (including CI tools) might install a version different from the one that other maintainers have installed and possibly fail unexpectedly. 
Besides this also prevents from using [`pnpm` `--frozen-lockfile`](https://pnpm.io/cli/install#--frozen-lockfile) which ensures that dependency specified in the lock files are installed.
I would consider removing it from the `.gitignore`

## `Node` and `pnpm` version are not aligned on development env and CI env

https://github.com/Valerioageno/tuono/blob/18a060e2bfcd20c06fb6dd6f9fe624a3a64b6b9c/package.json#L3

https://github.com/Valerioageno/tuono/blob/18a060e2bfcd20c06fb6dd6f9fe624a3a64b6b9c/.nvmrc#L1

While on workflows files a different version is used:

https://github.com/Valerioageno/tuono/blob/18a060e2bfcd20c06fb6dd6f9fe624a3a64b6b9c/.github/workflows/typescript.yml#L28

https://github.com/Valerioageno/tuono/blob/18a060e2bfcd20c06fb6dd6f9fe624a3a64b6b9c/.github/workflows/typescript.yml#L31

https://github.com/Valerioageno/tuono/blob/18a060e2bfcd20c06fb6dd6f9fe624a3a64b6b9c/.github/workflows/release.yml#L19-L23

## Possible solution to all above issues: reusable action for setup Node dependencies

Another improvement is to create a re-usable action to setup that reads configuration file to decide which tools version should be used:
- `setup-node` can detect the version from `.nvmrc`
- `pnpm/action-setup` already reads `package.json` `packageManager` field

<details>
<summary>Example of a setup nodeJS dependency action I'm using</summary>

```yml
name: Install Dependencies
description: This is a composite GitHub Action that sets up pnpm, node and installs the project's dependencies.

inputs:
  node-version:
    description: 'Explicit node version. Otherwise fallback reading `.nvmrc`'

runs:
  using: composite
  steps:
    - name: Install pnpm
      uses: pnpm/action-setup@v4

    - name: Setup Node.js (via input)
      if: ${{ inputs.node-version }}
      uses: actions/setup-node@v4
      with:
        node-version: ${{ inputs.node-version }}
        cache: 'pnpm'

    - name: Setup Node.js (via .nvmrc)
      if: ${{ !inputs.node-version }}
      uses: actions/setup-node@v4
      with:
        node-version-file: '.nvmrc'
        cache: 'pnpm'

    - name: Install Dependencies
      shell: bash
      run: pnpm install --frozen-lockfile

```

Usage:

```yml
     - name: Install Dependencies
       uses: ./.github/actions/install-dependencies
```

</details>

---

> [!TIP]
> As always I'm willing to setup a PR with all proposed changes!

marcalexiei:
## Context & Description

Close #97.

Basically all proposition of the issue should implemented here!

I have executed a `pnpm i --ignore-packages` in each example folder.

### DIFF
--- a/.github/workflows/deploy-documentation.yml
+++ b/.github/workflows/deploy-documentation.yml
@@ -15,27 +15,17 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
     steps:
-      - name: Check out code
+      - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 2
 
-      - name: Setup Node.js environment
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
+      - name: Install NodeJS Dependencies
+        uses: ./.github/actions/install-node-dependencies
 
       - name: Setup rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Install tuono
         run: cargo install tuono@0.12.3
-
-      - name: Install pnpm
-        run: npm i -g pnpm
-
-      - name: Install dependencies
-        run: pnpm install
 
       - name: Build project
         working-directory: ./apps/documentation
--- a/.github/actions/install-node-dependencies/action.yml
+++ b/.github/actions/install-node-dependencies/action.yml
@@ -1,0 +1,37 @@
+name: Install NodeJS Dependencies
+description: This is a composite GitHub Action that sets up pnpm, node and installs the project's dependencies.
+
+inputs:
+  node-version:
+    description: 'Explicit node version. Otherwise fallback reading `.nvmrc`. Use in conjunction with matrix'
+    required: false
+  registry-url:
+    description: 'https://github.com/actions/setup-node?tab=readme-ov-file#usage'
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: Install pnpm
+      uses: pnpm/action-setup@v4
+
+    - name: Setup Node.js (via input)
+      if: ${{ inputs.node-version }}
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ inputs.node-version }}
+        registry-url: ${{ inputs.registry-url }}
+        cache: 'pnpm'
+
+    - name: Setup Node.js (via .nvmrc)
+      if: ${{ !inputs.node-version }}
+      uses: actions/setup-node@v4
+      with:
+        node-version-file: '.nvmrc'
+        registry-url: ${{ inputs.registry-url }}
+        cache: 'pnpm'
+
+    - name: Install Dependencies
+      shell: bash
+      run: pnpm install --frozen-lockfile
--- a/.github/workflows/ci-documentation.yml
+++ b/.github/workflows/ci-documentation.yml
@@ -13,19 +13,16 @@ on:
 jobs:
   build:
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    name: Check build documentation
     timeout-minutes: 15
     runs-on: ubuntu-latest
 
     steps:
-      - name: Check out code
+      - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 2
 
-      - name: Setup Node.js environment
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
+      - name: Install NodeJS Dependencies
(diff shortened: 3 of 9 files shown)


.github/workflows/deploy-documentation.yml
```diff
@@ -15,28 +15,18 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
     steps:
-      - name: Check out code
+      - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 2
```
Valerioageno, line 21: Shouldn't we keed the `fetch-depth`? This might speed up a little bit the pipeline isn't it?

examples/mdx/pnpm-lock.yaml
Valerioageno, line 1: I wouldn't add this to the examples. The `tuono new [NAME]` command basically clones these folders on the user environment that might not wan't to use `pnpm` at all. 

For the pipelinse also we don't have anything that runs on the `examples/` folder so I don't think we have a real benefit. WDYT?

Valerioageno:
No `packages/` related pipeline got triggered in the PR. It doesn't look like but could it be related to these changes?

.github/workflows/deploy-documentation.yml
marcalexiei: You are right, I'll restore it!

examples/mdx/pnpm-lock.yaml
marcalexiei: I haven't thought about it. Better exclude `pnpm-lock` for `examples` folders then.

marcalexiei:
> It doesn't look like but could it be related to these changes?

It's strange the only relevant change is that I added a re-usable workflow. 
What is the repository settings under "Actions permissions" in [Actions general settings](https://github.com/Valerioageno/tuono/settings/actions)?

Valerioageno:
Great catch!

marcalexiei:
I had forgot adding a `./` in fronted of `.github` 😅
Better go to sleep and apply the rest of your review tomorrow

Valerioageno:
> I had forgot adding a `./` in fronted of `.github` 😅 Better go to sleep and apply the rest of your review tomorrow

😂

.github/workflows/deploy-documentation.yml
marcalexiei: How about keeping the default value, which is `1`?

From [@actions/checkout documentation](https://github.com/actions/checkout?tab=readme-ov-file#usage):
> Number of commits to fetch. 0 indicates all history for all branches and tags.
> Default: 1

From the looks of it we don't need to fetch commits more than 1 commit. 
I don't see any operation requiring git history, but I might be wrong.

Valerioageno: Oh I didn't know that the default is `1`. That's even better!

Valerioageno (approved):

.github/workflows/deploy-documentation.yml
```diff
@@ -15,28 +15,18 @@ jobs:
...
 
-      - name: Install pnpm
-        run: npm i -g pnpm
-
-      - name: Install dependencies
-        run: pnpm install
-
       - name: Build project
         working-directory: ./apps/documentation
         run: tuono build --static
```
marcalexiei, line 32: Another mystery... this part seems to hang now.

I also noticed now that in some workflow there is installed a fixed version of `tuono`.
Is this expected or the last available version should be used?

Valerioageno: Yeah I noticed that. I think the best would be a script that automatically infer the used tuono's version (by reading Cargo.toml) file. For the documentation I think we can move to the just deployed `0.12.3`

Valerioageno: I'm opening a PR now for updating the documentation version

Valerioageno: Merged. Could you try to rebase?

marcalexiei: All green!

PR #109, opened 2024-11-16
fix: handle hyphen in route

chenhunghan:
## Context & Description

I got this exception when trying to use a hyphen in route. e.g. 

```
src/routes/openid-conntect/index.rs
```

Screenshot 2024-11-16 at 13 45 17

assuming the fix it to replacing `-` to `_` that makes compiler happy.

### DIFF
--- a/crates/tuono/src/route.rs
+++ b/crates/tuono/src/route.rs
@@ -30,8 +30,9 @@ impl AxumInfo {
         let module_import = module
             .as_str()
             .to_string()
-            .replace(['/', '-'], "_")
+            .replace('/', "_")
             .replace('.', "_dot_")
+            .replace('-', "_hyphen_")
             .to_lowercase();
 
         if axum_route.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/tuono/src/route.rs
```diff
@@ -30,7 +30,7 @@ impl AxumInfo {
         let module_import = module
             .as_str()
             .to_string()
-            .replace('/', "_")
+            .replace(['/', '-'], "_")
```
Valerioageno, line 33: We should avoid this. I think this means that: `route/subroute` and `route-subroute` would get mapped with the same value. I'd follow the same pattern as the `.`: `_hyphen_`

crates/tuono/src/route.rs
```diff
@@ -186,6 +186,19 @@ mod tests {
         assert_eq!(dyn_info.module_import, "dyn_posts");
     }
 
+    #[test]
+    fn should_handle_hyphen() {
```
Valerioageno, line 190: Thanks for the test. Right now we have already a test that check for the correct route parsing. in `src/app.rs`. The test name is `should_collect_routes`. Could you please just add the map there? I agree that the test should better apply for the `route.rs` file but I'd keep all together for now

Valerioageno:
Thanks for spotting this!

Valerioageno (approved):

PR #115, opened 2024-11-17
ci: add format check on repository root

### LINKED ISSUE #108: Repo maintenance: enforce format on all supported files
Right now some files (like `json`, `yml` and `yaml`) are not automatically checked and formatted.

`prettier` supports formatting for all those files.
I would like also adding `md` / `mdx` to the list.

How about adding formatting checks for all these files using `prettier`?

marcalexiei:
## Context & Description

Close #108.

I had to add another GitHub Actions workflow since this check doesn't apply strictly apply to `typescript` or `rust` codebases.


Valerioageno (approved):
Nice addition!

PR #118, opened 2024-11-18
Update watchexec to v5 and watchexec-signals to v4

myanvoos:
Related to Issue #34 

Previously updating watchexec to v5 leads to this error:

```
error[E0308]: mismatched types
  --> crates/tuono/src/watch.rs:90:46
   |
90 |         if action.signals().any(|sig| sig == Signal::Interrupt) {
   |                                       ---    ^^^^^^^^^^^^^^^^^ expected `watchexec_supervisor::Signal`, found `watchexec_signals::Signal`
```

This PR fixes the import issue by changing `use watchexec-signals::Signal` to `use watchexec_supervisor::Signal`.

It also updates watchexec-signals to v4 with no further errors.

### DIFF
--- a/crates/tuono/src/watch.rs
+++ b/crates/tuono/src/watch.rs
@@ -3,8 +3,8 @@ use watchexec_supervisor::command::{Command, Program};
 
 use miette::{IntoDiagnostic, Result};
 use watchexec::Watchexec;
+use watchexec_signals::Signal;
 use watchexec_supervisor::job::{start_job, Job};
-use watchexec_supervisor::Signal;
 
 use crate::mode::Mode;
 use crate::source_builder::bundle_axum_source;
--- a/crates/tuono/src/watch.rs
+++ b/crates/tuono/src/watch.rs
@@ -3,8 +3,8 @@ use watchexec_supervisor::command::{Command, Program};
 
 use miette::{IntoDiagnostic, Result};
 use watchexec::Watchexec;
-use watchexec_signals::Signal;
 use watchexec_supervisor::job::{start_job, Job};
+use watchexec_supervisor::Signal;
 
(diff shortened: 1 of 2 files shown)


Valerioageno:
Thanks for addressing this. On my side, it panics when I try to start the dev environment on any project of the repo.

Screenshot 2024-11-18 at 16 39 49

Valerioageno (changes requested):
Could you please check that the fixes work on the repositories example projects and documentation website?

crates/tuono/src/watch.rs
```diff
@@ -3,8 +3,8 @@ use watchexec_supervisor::command::{Command, Program};
 
 use miette::{IntoDiagnostic, Result};
 use watchexec::Watchexec;
-use watchexec_signals::Signal;
 use watchexec_supervisor::job::{start_job, Job};
+use watchexec_supervisor::Signal;
```
Valerioageno, line 7: Why this change? Technically, `watchexec_supervisor::Signal` is just a re-export from `watchexec_signals::Signal`

myanvoos: You're right - what happened was I updated `watchexec` to 5.0.0 first (while `watchexec_signals` was still on 3.0.0) and encountered the error mentioned in the PR description. Changing the import to `watchexec_supervisor::Signal` fixed it. But updating `watchexec_signals` to 4.0.0 fixes the error as well since `watchexec` 5.0.0 is dependent on it. 

I'll revert the import back to `watchexec_signals::Signal` in a later commit as it's no longer necessary

Valerioageno (approved):

Valerioageno:
Thanks @myanvoos for addressing this!

PR #125, opened 2024-11-20
chore: add `examples` folder to workspaces and make examples packages name unique

marcalexiei:
## Context & Description

[Followup of "Tokio-runtime-worker issue"  Discord thread](https://discord.com/channels/1307411506735874208/1308386648505061426/1308391966077227072)

This PR aims to handle the examples folder `node_modules` folder without using `—ignore-workspace`.

Adding them to `pnpm` workspace creates a node_modules folder with symbolic links that should prevent the Tokio-runtime-worker error when running `turbo` cli within these folders.

> [!TIP]
> This should completely remove the need of independent node installation across packages simplifying project setup procedure.

---

### Examples folder renames

Also examples folder have renamed 
- `examples/tuono` → `examples/tuono-app`
- `examples/tutorial` → `examples/tuono-tutorial`
- `examples/mdx` → `examples/with-mdx`

> [!NOTE]
> No problem changing them to something different. 
> My only concern is that they should be unique in the future scenario where they can be involved 
> inside a `turbo` pipeline.

According to the previous renames `crates/tuono/src/scaffold_project.rs#create_new_project` has been updated to `tuono-app` as default value for `template` CLI argument.


Valerioageno:
> No problem changing them to something different. My only concern is that they should be unique in the future scenario where they can be involved inside a turbo pipeline.

What is the future scenario you see with them involved in `turbo`?

marcalexiei:
- If we are going to create e set of linter rules (eslint or similar) and we would like to run them on the examples before publishing a release.
- we add some unit test inside the examples scenario and we want to run them together before a release.

Maybe this doesn't have sense since these are very simple examples but just to want to point this out 🙂.

Valerioageno:
Yeah, makes sense. We don't know yet but better safer than sorry either way!

Issue #123, opened 2024-11-19
[Feature request]: Better error handling when target port is already in use


myanvoos:
### Is your feature request related to a problem? Please describe.

Currently Tuono's default target port is 3000. If that port is already in use and we try to run `tuono dev` we get the following error:

![image](https://github.com/user-attachments/assets/b6f58849-45e8-44ea-9259-83a8527868ff)

### Describe the solution you'd like

Tuono can try continuously searching for the next available port to use and stopping when it finds the next available one, instead of throwing an error.

Valerioageno:
As a quick improvement, though, I'd gracefully close the `tuono` CLI process with a meaningful message in case one of the two ports are already taken.

myanvoos:
> Hey @myanvoos, thanks for raising this. Definitely, we should improve the way we handle such case.
> 
> > Note: actually during development there are two open ports: `3000` for the rust server + `3001` for the vite server.
> 
> Personally, I wouldn't leave to the host machine (or to tuono itself) the duty of finding an available port since there are too many processes involved which make it error-prone IMHO. I'd rather allow the user to specify the desired port(s) using the (not available yet) `tuono.config.ts` file or a flag in the CLI like `tuono dev --port 4000`.
> 
> WDYT?
> 
> Tagging here also @marcalexiei for visibility since we already started discussing a `tuono.config.ts` file.

I like the idea of a tuono.config.ts file! And the CLI flag too, although it might not be clear at first glance whether we're specifying the Rust server port or the Vite port. Maybe `tuono dev --port x` can specify the Rust port and as a convention, the Vite port can be at `x+1`?

I'll have a go at implementing the CLI closing process in my fork^^

Valerioageno:
> Maybe tuono dev --port x can specify the Rust port and as a convention, the Vite port can be at x+1?

Yes!

Waiting for you PR then! :)

Valerioageno:
Let's create also a discussion - maybe on Discord - about the new `tuono.config.ts` and let's define a priority based roadmap for it

PR #126, opened 2024-11-20
Better error handling for target ports conflict

myanvoos:
Related to #123 

This PR adds better error handling in case one or both of the target ports are unavailable for use.

Previously the use of `unwrap()` causes tokio to panic if it cannot bind to port 3000, when the port is already being used by another process.

`let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();`

causes

```
file:///home/fourier/Documents/tutorial/node_modules/vite/dist/node/chunks/dep-CB_7IfJ-.js:45595
          reject(new Error(`Port ${port} is already in use`));
                 ^

Error: Port 3001 is already in use
    at Server.onError (file:///home/fourier/Documents/tutorial/node_modules/vite/dist/node/chunks/dep-CB_7IfJ-.js:45595:18)
    at Server.emit (node:events:518:28)
    at emitErrorNT (node:net:1943:8)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)

Node.js v22.11.0
thread 'tokio-runtime-worker' panicked at /home/fourier/.cargo/registry/src/index.crates.io-6f17d22bba15001f/watchexec-supervisor-3.0.0/src/job/task.rs:54:17:
all branches are disabled and there is no else branch
```

Now the output will simply be

```
  ⚡ Tuono v0.13.1

  Error: Failed to bind to either port 3000 or port 3001.
  Please ensure that ports 3000 and 3001 are not already in use by another process or application.
```

### DIFF
--- a/crates/tuono_lib/src/server.rs
+++ b/crates/tuono_lib/src/server.rs
@@ -32,16 +32,37 @@ impl Server {
     }
 
     pub async fn start(&self) {
-        let addr = "0.0.0.0:3000";
-        let listener = match tokio::net::TcpListener::bind(addr).await {
-            Ok(listener) => listener,
-            Err(e) => {
-                eprintln!("\n  Error: Failed to bind to address {}. {}", addr, e);
-                eprintln!("  Please ensure that port 3000 is not already in use by another process or application.");
-                std::process::exit(1);
-            }
-        };
+        let rust_addr = "0.0.0.0:3000";
+        let vite_addr = "0.0.0.0:3001";
 
+        // Get the port from the address to be displayed in the console
+        let rust_port = rust_addr.split(":")
+            .last()
+            .unwrap_or("unknown");
+        let vite_port = vite_addr.split(":")
+            .last()
+            .unwrap_or("unknown");
+
+        let rust_listener = tokio::net::TcpListener::bind(rust_addr).await;
+        let vite_listener = tokio::net::TcpListener::bind(vite_addr).await;
+
+        match (rust_listener, vite_listener) {
+            (Ok(rust_listener), Ok(_vite_listener)) => {
+                println!("\n  Using port {} for Rust server.", rust_port.bold());
+                println!("  Using port {} for Vite server.", vite_port.bold());
+
+                self.serve(rust_listener).await;
+            }
+            (Err(_listener), _) | (_, Err(_listener)) => {
+                eprintln!("\n  Error: Failed to bind to either port {} or port {}.", rust_port, vite_port);
+                eprintln!("  Please ensure that ports {} and {} are not already in use by another process or application.", rust_port, vite_port);
+                
+                std::process::exit(1);
+            }                            
+        }
+    }
+
+    pub async fn serve(&self, listener: tokio::net::TcpListener) {
         if self.mode == Mode::Dev {
             println!("  Ready at: {}\n", "http://localhost:3000".blue().bold());
             let router = self
--- a/crates/tuono_lib/src/server.rs
+++ b/crates/tuono_lib/src/server.rs
@@ -13,6 +13,15 @@ use crate::{
 const DEV_PUBLIC_DIR: &str = "public";
 const PROD_PUBLIC_DIR: &str = "out/client";
 
+fn extract_port(addr: &str) -> &str {
+    addr.split(":")
+        .last()
+        .unwrap_or_else(|| {
+            eprintln!("  Error: Failed to extract port from address {}", addr);
+            std::process::exit(1);
+        })
+}
+
 pub struct Server {
     router: Router,
     mode: Mode,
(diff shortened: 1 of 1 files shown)


Valerioageno:
I noticed now that I left `⚡ Tuono v0.13.1` in the source build. I think we can put also this in the CLI. I can do it in a separate PR

Valerioageno (changes requested):
Please move this logic into the `tuono` crate

crates/tuono/src/main.rs
```diff
@@ -1,5 +1,6 @@
 use tuono::cli::app;
 
-fn main() {
-    app().expect("Failed to start the CLI")
+#[tokio::main]
```
Valerioageno, line 3: Why the whole CLI has to be async?

crates/tuono/src/cli.rs
```diff
@@ -47,15 +47,58 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<()> {
...
+        std::process::exit(1);
+    })
+}
+
+async fn check_ports(mode: Mode) -> std::io::Result<()> {
+    println!("\nChecking ports...");
+
+    let rust_addr = "0.0.0.0:3000";
+    let rust_port = extract_port(rust_addr);
+    let rust_listener = tokio::net::TcpListener::bind(rust_addr).await;
```
Valerioageno, line 62: Can't we use `std::net::TcpListener` to make it blocking?

crates/tuono/src/cli.rs
```diff
@@ -47,15 +47,58 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<()> {
...
-pub fn app() -> std::io::Result<()> {
+fn extract_port(addr: &str) -> &str {
+    addr.split(":").last().unwrap_or_else(|| {
+        eprintln!("Error: Failed to extract port from address {}", addr);
+        std::process::exit(1);
+    })
+}
+
+async fn check_ports(mode: Mode) -> std::io::Result<()> {
+    println!("\nChecking ports...");
```
Valerioageno, line 58: I'd avoid logging this. This might be something to log in the future when we will have a `--debug` flag for the CLI. This info doesn't give any value to the user IMHO

crates/tuono/src/cli.rs
```diff
@@ -47,15 +47,58 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<()> {
...
+
+    let rust_addr = "0.0.0.0:3000";
+    let rust_port = extract_port(rust_addr);
```
Valerioageno, line 61: I'd create two static constants 
```
const TUONO_PORT: u16 = 3000;
const VITE_PORT: u16 = 3001;
```
And then just bind the listener with `std::net::TcpListener::bind(format!("0.0.0.0:{TUONO_PORT}"))`

I don't think we need to extract it from the static `0.0.0.0:3000` address

crates/tuono/src/watch.rs
Valerioageno, line 1: I'd remove these changes. 
Let's keep the PRs short, addressing a single topic. 
We will see how to pass the port in separate MRs (there are plenty of places in which we have to pass them down).

Valerioageno (changes requested):

crates/tuono/src/cli.rs
```diff
@@ -47,15 +47,58 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<()> {
...
         Actions::Dev => {
+            check_ports(Mode::Dev).await?;
+
             init_tuono_folder(Mode::Dev)?;
-            watch::watch().unwrap();
+
+            watch::watch().await.unwrap();
         }
         Actions::Build { ssg } => {
+            check_ports(Mode::Prod).await?;
```
Valerioageno, line 100: The port actually needs to be checked just when we do the `ssg` build. Could you please move this check after 

```rs
if ssg {
   // here
}
```

crates/tuono/src/cli.rs
```diff
@@ -47,12 +47,46 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<()> {
     Ok(())
 }
 
+fn check_ports(mode: Mode) -> std::io::Result<()> {
+    const TUONO_PORT: u16 = 3000;
```
Valerioageno, line 51: I'd put these two outside the function. Right below the `use` statements.

crates/tuono/src/cli.rs
```diff
@@ -47,12 +47,46 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<()> {
     Ok(())
 }
 
+fn check_ports(mode: Mode) -> std::io::Result<()> {
```
Valerioageno, line 50: Actually, this function doesn't return anything.

```suggestion
fn check_ports(mode: Mode) {
```

Valerioageno (changes requested):

PR #129, opened 2024-11-20
Add support for `api/` folder

Valerioageno:
## Context & Description

This PR adds the support to the `api/` folder in which the user can define any API endpoint that doesn't return any server side rendered HTML.


Valerioageno:
TODO. Add to this PR tests and write documentation about

PR #131, opened 2024-11-21
fix: prevent main thread from panicking when trying to overwrite an existing dir

myanvoos:
## Context & Description

This PR adds a directory existence check in the crate `tuono/scaffold_project.rs` to prevent the main thread from panicking when a directory already exists.

Previously running `tuono new [existing_dir]` gives

```
thread 'main' panicked at crates/tuono/src/scaffold_project.rs:72:29:
called `Result::unwrap()` on an `Err` value: Os { code: 17, kind: AlreadyExists, message: "File exists" }
```

Now it will just output `Error: Directory '[existing_dir]' already exists`.

### DIFF
--- a/examples/tuono-app/src/routes/api/health_check.rs
+++ b/examples/tuono-app/src/routes/api/health_check.rs
@@ -1,5 +1,5 @@
-use tuono_lib::axum::http::StatusCode;
 use tuono_lib::Request;
+use tuono_lib::axum::http::StatusCode;
 
 #[tuono_lib::api(GET)]
 pub async fn my_get_request(_req: Request) -> StatusCode {


Valerioageno:
Nice catch!

crates/tuono/src/scaffold_project.rs
```diff
@@ -68,6 +68,10 @@ pub fn create_new_project(folder_name: Option<String>, template: Option<String>)
     }
 
     if folder != "." {
+        if Path::new(&folder).exists() {
+            eprintln!("Error: Directory '{folder}' already exists");
+            return;
```
Valerioageno, line 73: How about triggering an exit code `std::process::exit(1)`?

myanvoos: > How about triggering an exit code `std::process::exit(1)`?

That can definitely be done. Do you think the template existence check

```rust
if new_project_files.is_empty() {
        println!("Template not found: {template}");
        return;
    }
```

should trigger an exit code as well? I did `return;` to keep it consistent with this check. Additionally there are some `unwrap()` calls that might be good to replace with error handling (can be in a separate PR)

myanvoos: > What do you think about also suggesting the right command? i.e. `Hint: you can scaffold a tuono project within an existing folder with "cd {folder} && tuono new ."`
> 
> Something like that

Yeah! I'll add it

For the template existence check maybe we can also suggest `Hint: available templates are: tuono-app, tuono-tutorial, with-mdx`

Valerioageno: Yeah, I think it is good to have OS meaningful exit codes. Unfortunately, in the early days, I focused more on functionality rather than quality. We should now start to improve the overall CLI quality.

For the hint, I'd link the user to the `examples/` folder, since the number of template might increase overtime.

PR #135, opened 2024-11-22
Better error handling when trying to overwrite an existing folder and using a non-existent template

myanvoos:
## Context & Description

This PR improves error handling for two cases:

- When attempting to overwrite an existing folder with `tuono new [existing_dir]`
- When attempting to use `tuono new` template flag with a non-existent template


Valerioageno (approved):
LGTM

PR #137, opened 2024-11-22
ci: test spellcheck via reviewdog

marcalexiei:


marcalexiei:
Now the actions runs correctly however is not able to report the issues found in the PR review:

> https://api.github.com/repos/Valerioageno/tuono/pulls/137/reviews: 403 Resource not accessible by integration []"
reviewdog: This GitHub Token doesn't have write permission of Review API [1],
so reviewdog will report results via logging command [2] and create annotations similar to

Even gratings `write-all` permissions seems to receive a 403 response status

marcalexiei (commented):
When PR comes from a forked repository with a `pull_request` event no permission can be set higher than `read`.
See [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) for more information.

This makes impossible to the review dog action to add review comments to the pull request.
This can happen only if the PR comes from within the repository: https://github.com/Valerioageno/tuono/pull/49.

---

At the moment I changed the workflow to use `GITHUB_TOKEN` and all the errors are reported using annotations.

At least now the language tool run is executed and issues are reported via error annotations: https://github.com/Valerioageno/tuono/actions/runs/11978476234

Unfortunately I was unable find a way to make the workflow fail if there is at least one error annotation.

---

### What about using a Personal access token as `github-token` input of `reviewdog/action-languagetool`?

It won't work. See [Using secrets in a workflow](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-secrets-in-a-workflow)

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

marcalexiei (commented):
I found a workaround:

```text
# If the workflow is triggered by a push use check reporter
#
# If the workflow is triggered by a PR coming from the
# same repository, the author has write access so we can
# use pr-review reporter since the GITHUB_TOKEN will have review
# write permissions
#
# Otherwise use annotation reporter that do not require any
# particular permission
```

WDYT?

Valerioageno (approved):
Thanks for addressing this! Let's see further PRs!

marcalexiei:
Looks like they work: https://github.com/Valerioageno/tuono/actions/runs/11987621865/job/33421986224?pr=141

PR #146, opened 2024-11-24
test(crates/tuono): Adding trycmd and a test for tuono help.

KingTheSim:
…or tuono_help.

## Context & Description

Introducing the trycmd crate and a test for the tuono --help command. Will continue to work on making a test for tuono build and the rest of the commands.


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
Valerioageno, line 1: We should exclude this file from the PR

crates/tuono/Cargo.toml
```diff
@@ -21,6 +21,7 @@ path = "src/lib.rs"
 
 [dependencies]
 clap = { version = "4.5.4", features = ["derive", "cargo"] }
+tempfile = "3.14.0"
```
Valerioageno, line 24: `tempfile` should be added to the [dev-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies)

crates/tuono/tests/cli_tests.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    assert!(
+        temp_main_rs_content.contains(MAIN_RS_CONTENT),
+        "'.tuono/main.rs' doesn't contain the expected code."
+    );
+}
+
+// Tests the 'tuono new' command by confirming that index.rs is build and that it contains the start of its function.
+#[test]
+fn test_tuono_new_success() {
```
Valerioageno, line 39: Let's avoid to test the `tuono new` command. It is too dependant to http requests right now. Better to handling it with a proper e2e test in a separate pipeline

crates/tuono/tests/cli_tests.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use tempfile::tempdir;
+
+const MAIN_RS_CONTENT: &str = "async fn main() {";
+const INDEX_RS_CONTENT: &str = "async fn get_server_side_props(_req: Request) -> Response {";
+
+// Tests the 'tuono build' command by confirming that main.rs is build and that it contains the start of main() function.
+#[test]
+fn test_tuono_build_success() {
+    let temp_dir = tempdir().expect("Failed to create a temporary test directory.");
```
Valerioageno, line 12: This test will always fail because the `tempdir` you created is always empty. You should create a util function like `fn create_temp_tuono_project() -> TempDir {}` that will be used across all the tesst and that setups a valid tuono project.

crates/tuono/tests/cli_tests.rs
KingTheSim: Should it be commented out or cut out as a whole?

.vscode/settings.json
marcalexiei: After removing .vscode folder, could you please add it to the `.gitignore` file?
It's better to avoid that IDE configuration files are committed accidentally 🥲

marcalexiei (commented):
We have introduced CI test also for `windows` and `macos` OS via #158.
Could you please try pulling those changes to see if the test are green also on them?

PR #147, opened 2024-11-24
chore(deps-dev): update `eslint` to v9

marcalexiei:
## Context & Description

Start working on `ESLint` update to v9.

With this update I'm also planning to remove more `any` as possibile from the typescript part


packages/router/src/components/RouterProvider.tsx
```diff
@@ -27,7 +23,11 @@ function RouterContextProvider({
...
     ...router.options,
     ...rest,
+    // @ts-expect-error router options do not have context as property and I was unable to find any usage of it,
+    //  so is this spread required?
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     context: {
+      // @ts-expect-error see previous
       ...router.options.context,
     },
   })
```
marcalexiei, line 33:
```suggestion
  router.update({
    ...router.options,
    ...rest,
  })
```

packages/router/src/globals.ts
```diff
@@ -0,0 +1,14 @@
+import type React from 'react'
+
+import type { Router } from './router'
+
+declare global {
+  interface Window {
+    __TSR__ROUTER__: Router
```
Valerioageno, line 7: This is something I forgot to rename when I was studying Tanstack router. I'd call this `__TUONO_ROUTER__` and also change the property setter down below

Valerioageno:
That is a great improvement!

packages/router/src/components/RouterProvider.tsx
Valerioageno: Yeah, there might be things left while I was investigating Tanstack router. Good to me

Valerioageno:
I'm going to log a task to remove all the things not used in the router

Valerioageno (approved):
LGTM

packages/router/src/components/RouterProvider.tsx
marcalexiei: As per [comment below](https://github.com/tuono-labs/tuono/pull/147#issuecomment-2509196235) this will be handled in a separate task

PR #153, opened 2024-11-25
feat: add `tuono.config.ts` support

### LINKED ISSUE #114: [Bug]: Documentation website loads all the @tabler/icon-react chunks
### Description

With the `tuono dev` command the package `@tabler/icon-react` loads all the chunks at every page load. It is a known bug on tabler side https://github.com/tabler/tabler-icons/issues/1233

### Expected behaviour

We should prevent to load all the icons

### How to reproduce

Just install the dependencies and run `tuono dev` and check the network tab

### Screenshots

Screenshot 2024-11-17 124729

### Tuono version

_No response_

### OS

_No response_

### Browser

_No response_

### Node version

_No response_

### Rust version

_No response_

### Crate version

_No response_

### Node Package Manger version

_No response_

### Additional context

There is a workaround that is not applicable for tuono because there is no interface available for tweaking the vite configuration on user side. We should consider to add this feature within a possible `tuono.config.ts` file soon!

Valerioageno:
## Context & Description

This PR is the first iteration for creating the main configuration file `tuono.config.ts`.

We are creating the binding for adding the alias option to the vite configuration.
Enable the ability to resolve #114.

## How does it work?
<details>
<summary>Original implementation</summary >
When the `tuono.config.ts` file exists, we build it into a commonjs module inside the `.tuono/config/config.js` file. Then any vite script before starting read the file and apply the changes.

I decided to create the folder `.tuono/config/` because when we will have to share the config with rust we will need to strip out all the ineligible values for the V8 engine so we will need to create a sibling `config-server.js` file (built again with vite and maybe a proper plugin).

Since the file is in the `./tuono/config` folder using `__dirname` will cause issues with the path resolution so, in order to overcome this, I created a `path` module that has the `path.join()` function that automatically strips out the wrong filepath and correct `__dirname` with the tuono folder 
</details>

[New proposed approach can be found in the comment below](https://github.com/Valerioageno/tuono/pull/153#issuecomment-2508142877)


packages/tuono/src/config/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface TuonoConfig {
```
marcalexiei, line 1: To mimic `vite` structure I would consider to name this `TuonoUserConfig`.
Maybe in the future we need to perform some normalization and we mind end up to having a `TuonoNormalizedConfig` or something similar.

packages/tuono/src/config/types.ts
```diff
@@ -0,0 +1,5 @@
...
+  vite?: {
+    alias?: Record<string, string>
```
marcalexiei, line 3: Since we are passing this option directly to `vite` we might consider to extract this types from the `vite` config type itself rather than hardcoding them on our side

packages/tuono/src/config/types.ts
marcalexiei, line 1: How about adding a `defineConfig` or `defineTuonoConfig` helper?
User should be already familiar with this approach from `vite` ecosystem configs.
Having an helper function is also used by [`typescript-eslint`](https://typescript-eslint.io/getting-started#step-2-configuration): `tseslint.config`

packages/tuono/src/build/index.ts
marcalexiei, line 1: There  is a lot of config merging happing here!
How about using `mergeConfig` helper from `vite` when merging `BASE_CONFIG` with other configs?

---

Another thought for future reference:
how about adding some validation to the user provided config using library like [zod](https://zod.dev). 

Right now we don't have many options but this might change in the future. What do you think?

packages/tuono/src/config/types.ts
Valerioageno: I thought to it but this would possible complicate the way we will trad this config from the RUST CLI... better to keep it as object for now in my opinion

packages/tuono/src/build/index.ts
Valerioageno: I think it would be cool when we will have some strict validation. For now all the config are optional so it sounds overkill now

packages/tuono/src/config/types.ts
Valerioageno: I don't like it, honestly. I prefer just `TuonoConfig` like `NextConfig`

Valerioageno:
@marcalexiei The night will bring suggestions (😄).

I think the best solution would be to create a util function that just iterate over the config strings and replace any occurence of `.tuono/config/` with an empty string and get rid of the super-weird `import { path } from 'tuono/config'`. 

This would be still a bit error-prone for cases like: 
```
{
    [__dirname]: 'whatever'
}
```
Because I don't see the point of changing the keys of the object, but now that I'm writing it sounds even less possible.

marcalexiei (commented):
> I think the best solution would be to create a util function that just iterate over the config strings and replace any occurence of .tuono/config/ with an empty string

After a quick review I could't think of an alternative, maybe we could play with `path.relative` and other utils to make this more transparent as possible for the user.

I'm with you by trusting the Night to bring clever suggestions!

---

> get rid of the super-weird import { path } from 'tuono/config'.

I agree with you, this is super weird and I'm not very comfortable in using something like this. 
Maybe we could give it another name but surely it remains error prone.
Users might just copy vite config path related properties here, the `vite` wrapper clearly doesn't help 😂.

---

Valerioageno:
Decision after brainmelting:

No bundler needed for the build of the `tuono.config.ts`. It will be built into `.tuono/config/config.mjs`.
The paths replacement will work (for now) only in the alias key and they will:
- If the path is relative turn them into an absolute path from the project root
- If the path is absolute replace `.tuono/config/` with `/`

marcalexiei (commented):
I applied all the changes we discussed to day and added relevant unit tests.
Let me know if everything checks out also on your end.

marcalexiei (approved):

Issue #150, opened 2024-11-25
[Bug]: Error when running tuono run dev


adityapandeydev:
### Description

I am not yet sure if the problem is my system or not, but I thought the best place to know will be here. I am sorry if this was a time waste and I will delete the issue. But the problem I am facing is upon running tuono dev (I ran the command just after doing npm install without making any changes in the default code) I am getting this error (I have uploaded the pictures).
![image](https://github.com/user-attachments/assets/7e010be0-530c-4376-a44b-5fde7e47f02e)
![image](https://github.com/user-attachments/assets/a938af14-d93a-4b4b-b906-699dfe5ae35f)

### Expected behaviour

I just expected to see it working, like you normally run npm run dev just after creating a new project.

### How to reproduce

I ran the tuono dev command immediately after creating the project and running the npm install command.

### Tuono version

0.14.1

### OS

Windows

### Node version

20.13.1

### Rust version

1.79.0

### Node Package Manger version

10.8.3

Valerioageno:
Looks like you don't have installed the npm dependencies. What node package manager are you using?

Valerioageno:
Thanks for reporting. That definitely is a bug with Windows.

Valerioageno:
Hey @adityapandeydev, thanks for reporting it. The fix has just been published. You can update `tuono`by running

```sh
cargo install tuono@0.14.3
```

Let me know if you will find something broken 😄

adityapandeydev:
Unfortunately I am still finding the same error. I installed the updated version as you said.
![image](https://github.com/user-attachments/assets/c4417025-9960-40a8-8f0b-3bfb473476d7)
![image](https://github.com/user-attachments/assets/7f3eb9d0-b312-4161-b722-9e51a4bee961)
![image](https://github.com/user-attachments/assets/fee81c7d-83e2-4820-aab2-1eb8c053f500)
Also the name of the project might be the same as before but I deleted the project and created a new one with the same name, so all the node modules and everything is fresh.
I also tried to create a new project in a different directory to see if this might be a caching issue, i still faced the same error but made a new discovery (don't know if you already knew this) but when this error comes the localhost port is getting occupied, because when I created the project in the new directory and ran the tuono dev command I was getting a different error which was [port 3001 is already in use], I had to ctrl + c the terminal of the previous project and then when I again ran I started getting the same error.

Valerioageno:
Mmmm. This seems a problem with the V8 build...

What is your PC cpu?

[...]

PR #158, opened 2024-11-26
ci: add `windows-latest` to test matrix for both `rust` and `typescript`

marcalexiei:
## Context & Description

Related to #150.

### Additional changes:

- `package.json` `report:format` scripts were incorrect
- make jobs order consistent between `rust` and `typescript` workflows

### DIFF
--- a/.github/workflows/repo-root-ci.yml
+++ b/.github/workflows/repo-root-ci.yml
@@ -28,4 +28,4 @@ jobs:
         uses: ./.github/actions/install-node-dependencies
 
       - name: Test project
-        run: pnpm repo:root:format
+        run: pnpm repo:root:format:check
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,6 +14,31 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  build_and_test:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    name: Build and test rust crates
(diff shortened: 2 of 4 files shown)


marcalexiei (commented):
To resolve cross platform issues handling line ending characters I moved the expected results from versioned files (`packages/fs-router-vite-plugin/tests/generator/*/*.expected.ts`) to a snapshot file.

packages/fs-router-vite-plugin/tests/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,182 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`generator works > should wire-up the routes for a "mdx" tree 1`] = `
```
Valerioageno, line 3: Is it possible to split this file into multiple cases? I wouldn't have a huge assertion file

marcalexiei: Would you prefer something like this https://vitest.dev/guide/snapshot.html#file-snapshots?

Valerioageno: Yeah. Every test case its own snap file

marcalexiei: Updated `fs-router-vite-plugin/tests/generator.spec.ts` to use `toMatchFileSnapshot`

PR #166, opened 2024-11-26
ci: trigger workflow only on `push` only on `main`

marcalexiei:
## Context & Description

`push` should trigger CI only when performed on  `main` branch

> [!NOTE] 
> CI is still executed on every push on a branch with an opened PR.
> (The PR target can be any branch)


Valerioageno (approved):
LGTM

PR #172, opened 2024-11-27
chore(renovate): enable schedule

marcalexiei:
## Context & Description

Set renovate schedule to `every 2 weeks on monday` to reduce PR frequency.


Valerioageno (approved):
Cool!

PR #174, opened 2024-11-27
ci(rust): avoid `beta` and `nightly` toolchains to be executed on macos and windows

marcalexiei:
## Context & Description

On  `rust` CI remove `nightly` and `beta` toolchains from `windows` and `macOS`


Valerioageno (approved):
LGTM

PR #196, opened 2024-12-03
refactor: revamp documentation

Valerioageno:


marcalexiei (commented):
I like the new layout of the side navigation. Here are a few suggestions that may render it even better:

### 1. I would consider moving "Contributing" to the bottom of the list with a separator.  

The user does not need to access the page very often. Actually, I think they need to access it once.
It might be joined in the future with another page about project maintenance (e.g: how to perform a release)

### 2. Add subtitles to each of relevant section to better identify its content.

Since we are using a visual separator, I would also consider adding a title to clearly identify what items are included in that section.

> [!NOTE]
> Maybe some pages under "Overview" can be moved inside "API reference", but we should consider this after adding the content to the various pages.

## Navigation updates proposal

1. Untitled
   - Home
   - Getting started
   - Tutorial
2. Overview
   - Core Concepts 
   - Routing
   - Rendering
   - Styles
   - Integrations
3. "API reference" or just "References"
   - Components
   - Hooks
5. Untitled
   - Contributing

Valerioageno:
I like it! Let's do it

marcalexiei (commented):
Crawled Discord server to retrieve useful messages with content that can be added to the documentation:

- [Rust crates overview](https://discord.com/channels/1307411506735874208/1309985604355887225/1309987369574989855)
- [Explanation about build files](https://discord.com/channels/1307411506735874208/1314506155355017246/1314512901255462974)
- (Adding it for completeness) [Implementation breakdown dev.to article](https://discord.com/channels/1307411506735874208/1307411506735874211/1313192044562219040)

---

These are all the useful message I was able to find so far, I hope they help!

---

Other missing docs part:
- [cookie documentation](https://github.com/tuono-labs/tuono/pull/201)

Issue #190, opened 2024-12-01
Add cookies to Response


Valerioageno:
All the cookies sent from the client to the server are already included in the `Request.headers` struct. We won't create any layer on top of that in order to let the users decide which library to use to read the cookies.

With this task, we add the cookies to the Response so that the user can define server sent cookies.

[axum-extra cookie](https://docs.rs/axum-extra/latest/axum_extra/extract/cookie/index.html) crate.

PR #201, opened 2024-12-06
feat: add cookies to the response

Valerioageno:
## Context & Description

This PR adds the support for reading and passing cookies to the HTTP response.

Passing cookies to the response is now possible by:

```rust
let mut props = Props::new(data);
props.add_cookie(Cookie::new());

return Response::Props(props);
```

The `Cookie` stuct has multiple implementation, so it is also possible to use different syntax like: `props.add_cookie(("name", "value"))`.

Also, `tuono_lib` re-export the `axum-extra::extractor::cookie` lib so that the end user can manage both `CookieJar` and `Cookie` structs

```rust
use tuono_lib::cookie::*;
```

Since the new support for cookies is now also possible to interact with the cookies passed from the `Request` like:

```rust
fn handler(req: Request) -> Response {
     let cookies = CookieJar::from_headers(&req.headers);
}
```

This PR closes https://github.com/tuono-labs/tuono/issues/190

### DIFF
--- a/crates/tuono_lib/Cargo.toml
+++ b/crates/tuono_lib/Cargo.toml
@@ -19,6 +19,7 @@ include = [
 [dependencies]
 ssr_rs = "0.7.0"
 axum = {version = "0.7.5", features = ["json", "ws"]}
+axum-extra = {version = "0.9.6", features = ["cookie"]}
 tokio = { version = "1.37.0", features = ["full"] }
 serde = { version = "1.0.202", features = ["derive"] }
 erased-serde = "0.4.5"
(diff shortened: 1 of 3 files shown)


marcalexiei (commented):
Looks good to me!

My two cents: the lack of unit tests. 
Is it possible to add them to this module, or is it something non-trivial?
The functionality is built upon another library but I think we should start adding unit test whenever is possible.

----

Added a reference to this PR in the documentation revamp to remember to add cookies reference there:
- https://github.com/tuono-labs/tuono/pull/196#pullrequestreview-2485694361

Valerioageno:
Definitely non-trivial. But you raised a very good point. Better start to create a suite for it. I'll focus on it in a separate task

Valerioageno:
Logged here! https://github.com/tuono-labs/tuono/issues/203

Valerioageno:
Sorry. I created a task for the `integration tests` since are non-trivial. I also added a `unit test` as you suggested!!

PR #205, opened 2024-12-07
refactor(typescript): use generic instead of simple array type

marcalexiei:
## Context & Description

Switch [`@typescript-eslint/array-type`](https://typescript-eslint.io/rules/array-type/) rule to `generic`.

```ts
type MyArray = string[]
// now should be
type MyArray = Array<string>
```

A more detailed explanation can be found on the following blog post:
https://tkdodo.eu/blog/array-types-in-type-script


Valerioageno (approved):
Finally! Thanks

PR #211, opened 2024-12-07
docs: add link and navigation page content

Valerioageno:


marcalexiei (commented):
Due to the large amount of review dog comments I created different commits with my proposed changes.
Let me know what do you think about them.

----

When writing the section about the `useRouter` section I have some thought about 2 new possible features:

1. `useRouter` should also provide a `replace` method to handle redirects
2. Create declarative components to perform the same navigation operations?
     I would consider using something like <Navigate /> with an optional boolean parameter `replace` to decide if is a push or a replace operation.
     ```tsx
     if (sucesss) return <Navigate replace  to="...." />
      ```
      Maybe is overkill because a user could create it himself?

marcalexiei (approved):

PR #213, opened 2024-12-08
feat: add polyfill db to documentation

Valerioageno:
## Context & Description

This PR initialize a `polyfill-db.json` file in the documentation in order to keep track of the common missing polyfills


apps/documentation/public/polyfill-db.json
```diff
@@ -0,0 +1,11 @@
...
+		"polyfills": [
+				{ 
+						"identifier": "HTMLElement",
+						"imports": ["@lit-labs/ssr-dom-shim", "HTMLElement"]
+				},
+				{ 
+						"identifier": "customElements",
+						"imports": ["@lit-labs/ssr-dom-shim", "customElements"]
+				}
+}
```
marcalexiei:
```suggestion
				}
		]		
}
```

marcalexiei: The json is invalid, the polyfills array is not closed

apps/documentation/public/polyfill-db.json
```diff
@@ -0,0 +1,11 @@
+{
+		"polyfills": [
+				{ 
+						"identifier": "HTMLElement",
+						"imports": ["@lit-labs/ssr-dom-shim", "HTMLElement"]
```
marcalexiei: The names `imports` suggest a list however this array is referencing to a package name and a named export so IMHO is confusing.

```suggestion
             "import":  {  
						   "package": "@lit-labs/ssr-dom-shim", 
						   "name": "HTMLElement"   
						 }
```

marcalexiei: I'm wondering if we should support a list of available polyfills for each element.
Example accounting for my previous comment:

```json
{
  "polyfills": [
    {
      "identifier": "HTMLElement",
      "imports": [
        {
          "package": "@lit-labs/ssr-dom-shim",
          "name": "HTMLElement"
        },
        {
          "package": "@anotherOne",
          "name": "HTMLElement"
        }
      ]
    }
  ]
}

```

marcalexiei (commented):
1. Could this json be referenced from somewhere in the documentation to let the user know official polyfills (Overview -> Core concepts)?

----

### Additional change

Now prettier runs on all documentation files

Valerioageno (commented):
That"s why I waited to merge 😄Thanks

Valerioageno:
Yeah I thought to create a page in which the user can search for it. For now let's just create the db to keep track of them. In the close future we will also do that!

apps/documentation/public/polyfill-db.json
Valerioageno: I think it gets hard to maintain. WDYT?

marcalexiei: Ok, let's keep a single import for each polyfill.

Valerioageno: Yeah. Better to use `import`

Valerioageno (commented):
refactor: update 'imports' key to 'import'

PR #215, opened 2024-12-09
docs: optimize public images

marcalexiei:
## Context & Description

I ran [imageoptim.com](https://imageoptim.com/mac) on `apps/documentation/public/` images.
There should be no quality loss, on the contrary size reduction is significant.


Valerioageno (approved):
Thanks for this!

Issue #216, opened 2024-12-09
[Bug]: Tuono dev crashes once with tuono dynamic


pveierland:
### Description

I tested both `React.lazy` and `tuono.dynamic` with tuono 0.16.0 as described.

### Dynamic component

```
// DynamicComponent.tsx
import { type JSX } from "react";

export const DynamicComponent = (): JSX.Element => {
	return <>DynamicComponent</>;
};

export default DynamicComponent;
```

### Test 1: `React.lazy`

```
// testReactLazy.tsx

import React, { Suspense, type JSX } from "react";

const DynamicComponent = React.lazy(() => import("../components/DynamicComponent"));

const LoadingComponent = (): JSX.Element => {
	console.log("Loading");
	return <>Loading</>;
};

export default function IndexPage(): JSX.Element | null {
	return (
		<Suspense fallback={<LoadingComponent />}>
			<DynamicComponent />
		</Suspense>
	);
}
```

This approach works with `tuono dev`, but results in the following client side warning/error logged in console:

```Uncaught Error: The server could not finish this Suspense boundary, likely due to an error during server rendering. Switched to client rendering.```

### Test 2: `tuono.dynamic`

```
// testTuonoDynamic.tsx

import { Suspense, type JSX } from "react";
import { dynamic } from "tuono";

const DynamicComponent: any = dynamic(() => import("../components/DynamicComponent"));

const LoadingComponent = (): JSX.Element => {
	console.log("Loading");
	return <>Loading</>;
};

export default function IndexPage(): JSX.Element | null {
	return (
		<Suspense fallback={<LoadingComponent />}>
			<DynamicComponent />
		</Suspense>
	);
}
```

This approach crashes `tuono dev` when starting the server, but after resetting it works, however with the same client side warning/error logged in console:

```Uncaught Error: The server could not finish this Suspense boundary, likely due to an error during server rendering. Switched to client rendering.```

The following is output by `tuono dev` upon start:

```
file:///project/node_modules/tuono-lazy-fn-vite-plugin/dist/esm/index.js:29
      if (init.callee.name === TUONO_DYNAMIC_FN_ID) {
                      ^

[vite-plugin-tuono-lazy-loading] unknown file: Cannot read properties of undefined (reading 'name')
file: /project/src/routes/fresh.tsx
    at file:///project/node_modules/tuono-lazy-fn-vite-plugin/dist/esm/index.js:29:23
    at Array.forEach (<anonymous>)
[...]

PR #222, opened 2024-12-14
fix: update `lazyLoadComponent` fn name to `__tuono__internal__lazyLoadComponent`

Valerioageno:
## Context & Description

This PR is the first related to this issue (https://github.com/tuono-labs/tuono/issues/216) about fixing the `dynamic` fn to let it be used from the users.

See https://github.com/tuono-labs/tuono/issues/216#issuecomment-2529385045


.prettierignore
```diff
@@ -1 +1,2 @@
-pnpm-lock.yaml
\ No newline at end of file
+pnpm-lock.yaml
+packages/lazy-fn-vite-plugin/tests/sources/*
```
marcalexiei, line 2:
```suggestion
```

This has been already excluded in `packages/lazy-fn-vite-plugin/.prettierignore` so I think this can be removed from there.

packages/lazy-fn-vite-plugin/src/constants.ts
```diff
@@ -1,3 +1,3 @@
 export const TUONO_DYNAMIC_FN_ID = 'dynamic'
-export const TUONO_LAZY_FN_ID = 'lazyLoadComponent'
+export const TUONO_LAZY_FN_ID = 'tuono__internal__lazyLoadComponent'
```
marcalexiei, line 2:
```suggestion
export const TUONO_LAZY_FN_ID = '__tuono__internal__lazyLoadComponent'
```

I would consider using a `__` prefix to avoid suggestion when typing  `tuono` since this type is exported,

I have a remembrance that we have discussed this, but probably I have just imagined it 😅

packages/lazy-fn-vite-plugin/src/index.ts
```diff
@@ -33,7 +33,7 @@ const RemoveTuonoLazyImport: PluginItem = {
 
 /**
  * [CLIENT build]
- * This plugin replace the `dynamic` function with the `lazyLoadComponent` one
+ * This plugin replace the `dynamic` function with the `tuono__internal__lazyLoadComponent` one
```
marcalexiei, line 36:
```suggestion
 * This plugin replace the `dynamic` function with the `__tuono__internal__lazyLoadComponent` one
```

packages/lazy-fn-vite-plugin/tests/transpileSource.test.ts
```diff
@@ -1,27 +1,12 @@
+import fs from 'node:fs/promises'
+
+import path from 'node:path'
```
marcalexiei, line 3:
```suggestion
import fs from 'node:fs/promises'
import path from 'node:path'
```

packages/tuono/src/index.ts
```diff
@@ -7,7 +7,7 @@ export {
   Link,
   RouterProvider,
   dynamic,
-  lazyLoadComponent,
+  tuono__internal__lazyLoadComponent,
```
marcalexiei, line 10:
```suggestion
  __tuono__internal__lazyLoadComponent,
```

packages/router/src/index.ts
```diff
@@ -2,5 +2,5 @@ export { RouterProvider } from './components/RouterProvider'
 export { default as Link } from './components/Link'
 export { createRouter } from './router'
 export { createRoute, createRootRoute } from './route'
-export { dynamic, lazyLoadComponent } from './dynamic'
+export { dynamic, tuono__internal__lazyLoadComponent } from './dynamic'
```
marcalexiei, line 5:
```suggestion
export { dynamic, __tuono__internal__lazyLoadComponent } from './dynamic'
```

packages/router/src/dynamic.tsx
```diff
@@ -41,7 +41,9 @@ export const dynamic = (importFn: ImportFn): React.JSX.Element => {
   return <></>
 }
 
-export const lazyLoadComponent = (factory: ImportFn): RouteComponent => {
+export const tuono__internal__lazyLoadComponent = (
```
marcalexiei, line 44:
```suggestion
export const __tuono__internal__lazyLoadComponent = (
```

packages/lazy-fn-vite-plugin/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src", "tests", "vite.config.ts"]
+  "include": ["src", "tests/transpileSource.test.ts", "vite.config.ts"]
```
marcalexiei, line 3: This kind of tsconfig is error prone since user has to manually include each future tests file.

Adding a match to `exclude` option targetting new `tests/sources` folder avoids this scenario .

https://www.typescriptlang.org/tsconfig/#exclude

> Specifies an array of filenames or patterns that should be skipped when resolving [include](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#include).

```suggestion
  "include": ["src", "tests", "vite.config.ts"],
  "exclude": ["tests/sources"]
```

packages/lazy-fn-vite-plugin/tests/transpileSource.test.ts
```diff
@@ -32,46 +17,33 @@ type ViteTransformHandler = Exclude<
...
+describe('"dynamic" fn', async () => {
+  const folderNames = await fs.readdir(process.cwd() + '/tests/sources')
+
+  it.each(folderNames)(
+    'should correctly build the "%s" dynamic fn',
+    async (folderName) => {
+      const testDirPath = `${process.cwd()}/tests/sources/${folderName}`
+
+      const sourceBuf = await fs.readFile(path.join(testDirPath, 'source.tsx'))
+      const source = sourceBuf.toString()
```
marcalexiei, line 29: Specifying 'utf-8' as second `readFile` parameter directly returns a string, without the need to an additional variable.

```suggestion
      const source = await fs.readFile(
        path.join(testDirPath, 'source.tsx'),
        'utf-8',
      )
```

marcalexiei (changes requested):

.prettierignore
Valerioageno: I don't know the reason, but my IDE still complain without it

marcalexiei (approved):
I applied my proposed changes.
If everything's checks out for you we can merge this!

PR #223, opened 2024-12-15
fix(packages/lazy-fn-vite-plugin): make `dynamic` function detection stricter

marcalexiei:
## Context & Description

Fix issue 2 reported in https://github.com/tuono-labs/tuono/issues/216#issuecomment-2529385045

> unknown file: Cannot read properties of undefined (reading 'name') This error appear when dynamic is the only function exported by the tuono library import { dynamic } from 'tuono'. We need to improve the internal vite plugin + adding this test case

----

To fix the issue I replaced all casting inside `turnLazyIntoStatic` with helper function from `@babel/types` that provide check the correct node runtime and provide types accordingly. 
E.g.:
```tsx
isCallExpression(node: Node | null | undefined, opts?: Opts<CallExpression> | null): node is CallExpression;
```

I also refactored types by giving them name to reference `babel` or `vite` to avoid any confusion.

### DIFF
--- a/packages/lazy-fn-vite-plugin/src/index.ts
+++ b/packages/lazy-fn-vite-plugin/src/index.ts
@@ -1,7 +1,7 @@
 import { transformSync } from '@babel/core'
+import type { PluginItem as BabelPluginItem } from '@babel/core'
 import * as BabelTypes from '@babel/types'
 import type { Plugin as VitePlugin, Rollup } from 'vite'
-import type { PluginItem as BabelPluginItem } from '@babel/core'
 
 import {
   TUONO_MAIN_PACKAGE,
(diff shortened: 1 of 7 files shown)


packages/lazy-fn-vite-plugin/src/index.ts
```diff
@@ -91,28 +105,38 @@ const TurnLazyIntoStaticImport: PluginItem = {
...
     enforce: 'pre',
-    transform(code, _id, opts): string | undefined | null {
+    transform(code, _id, opts): Rollup.TransformResult {
+      /**
+       * @todo we should exclude non tsx files from this transformation
+       *       this might benefit build time avoiding running `includes` on non-tsx files.
+       *       This can be executed using `_id` parameter
+       *       which is the filepath that is being processed
+       */
+
```
marcalexiei, line 119: Let me know what do you think about this

packages/lazy-fn-vite-plugin/src/index.ts
```diff
@@ -91,28 +105,38 @@ const TurnLazyIntoStaticImport: PluginItem = {
...
+          ['@babel/plugin-syntax-typescript', { isTSX: true }],
+        ]
+
+        if (opts?.ssr) {
+          plugins.push(RemoveTuonoLazyImport, TurnLazyIntoStaticImport)
+        } else {
+          plugins.push(ReplaceTuonoLazyImport)
+        }
+
+        const res = transformSync(code, { plugins })
```
marcalexiei, line 135: I have removed `sourceMaps: true` since sourceMap are not used:
This function is only returning the code processed by babel (see below `return`)

packages/lazy-fn-vite-plugin/tests/sources/dynamic-only/server.expected.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { Suspense, type JSX } from "react";
+import "tuono";
```
marcalexiei, line 2: This should not be present
```suggestion
```

packages/lazy-fn-vite-plugin/src/index.ts
Valerioageno: I agree. We should be extra-careful though. We should support at least js/jsx/mdx

Valerioageno (approved):
LGTM. Thanks

packages/lazy-fn-vite-plugin/src/index.ts
marcalexiei: New separate issue: #224

PR #227, opened 2024-12-17
refactor(packages/lazy-fn-vite-plugin): add filter to avoid processing unrelated files

### LINKED ISSUE #224: feat(packages/lazy-fn-vite-plugin): apply transformation only for valid filles
https://github.com/tuono-labs/tuono/pull/223#discussion_r1885722802

> we should exclude non tsx files from this transformation
> this might benefit build time avoiding running `includes` on non-tsx files.
> This can be executed using `_id` parameter
> which is the filepath that is being processed

marcalexiei:
## Context & Description

Close #224

### DIFF
--- a/packages/tuono/src/build/index.ts
+++ b/packages/tuono/src/build/index.ts
@@ -17,6 +17,12 @@ const VITE_PORT = 3001
 function createBaseViteConfigFromTuonoConfig(
   tuonoConfig: TuonoConfig,
 ): InlineConfig {
+  /**
+   * @warning Keep in sync with {@link LazyLoadingPlugin} tests:
+   * packages/lazy-fn-vite-plugin/tests/transpileSource.test.ts
+   */
+  const pluginFilesInclude = /\.(jsx|js|mdx|md|tsx|ts)$/
+
   const viteBaseConfig: InlineConfig = {
     root: '.tuono',
     logLevel: 'silent',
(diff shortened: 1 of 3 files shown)


packages/tuono/src/build/index.ts
```diff
@@ -17,6 +17,12 @@ const VITE_PORT = 3001
 function createBaseViteConfigFromTuonoConfig(
   tuonoConfig: TuonoConfig,
 ): InlineConfig {
+  /**
+   * @warning Keep in sync with {@link LazyLoadingPlugin} tests:
+   * packages/lazy-fn-vite-plugin/tests/transpileSource.test.ts
+   */
+  const pluginFilesInclude = /\.(jsx|js|mdx|md|tsx|ts)$/
```
Valerioageno, line 24: Since we have a monorepo, what do you think about creating a `constants|shared|i-dont-know-yet` folder to manage all the shared things?

marcalexiei: This is the first time I find myself "duplicating"**\*** code in the repo.
Right now I'm unaware of other redundant / duplicated content, so, from my point of view,
creating a new package right now adds only an overhead.

If you are aware of other scenarios like this,
I suggest to open a separate issue where we list all those cases,
then we decide if a new package is worth it and, eventually, how to name it.

**\*** I use `"` because the regexp resides here and in a test file so it's not duplicated in user-facing code)

Valerioageno: Agree. LGTM!

Valerioageno (approved):

PR #228, opened 2024-12-17
chore(apps/documentation): remove `tsconfig.node.json`

marcalexiei:
## Context & Description

`apps/documentation` doesn't have a `vite.config.ts` file, hence is present is not required. 
We can safely remove it.


Valerioageno:
I tried already to remove it, but something was breaking. Have you checked that `tuono dev | build` work correctly?

marcalexiei:
It all works correctly from my side.
Right now there is no type checking involved during `tuono build | dev` or process.

Besides that, `apps/documentation/tsconfig.node.json` was targeting only a `vite.config.ts` file
which is not present inside `apps/documentation`. So I think we are good to go.

Valerioageno (approved):

Issue #226, opened 2024-12-17
[Bug]: Catch-all paths has double underscore causing build warning


pveierland:
### Description

The new catch-all path feature uses double underscores which triggered build warnings:

```
warning: module my_prefix__catch_all_inner_hyphen_store_hyphen_path should have a snake case name
  --> crates/project/.tuono/main.rs:15:25
   |
15 |                     mod my_prefix__catch_all_inner_hyphen_store_hyphen_path;
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: convert the identifier to snake case: nix_store_dyn_catch_all_inner_hyphen_store_hyphen_path
```

### Expected behaviour

No build warning.

### How to reproduce

Create `[...catch_all].{rs,tsx}` route.

Tuono version 0.16.3 (web) + 0.16.2 (CLI).

mustafasegf:
one workaround i can think of is to add `.replace("__", "_")` at the end of https://github.com/tuono-labs/tuono/blob/main/crates/tuono/src/route.rs#L56. if you ok with that I can make a pr for it

PR #229, opened 2024-12-18
fix: catch-all paths replace double underscore with single underscore

mustafasegf:
close https://github.com/tuono-labs/tuono/issues/226

replace double underscore with single underscore

### DIFF
--- a/crates/tuono/src/route.rs
+++ b/crates/tuono/src/route.rs
@@ -52,9 +52,8 @@ impl AxumInfo {
                     .replace('/', "_")
                     .replace('-', "_hyphen_")
                     .replace('[', "dyn_")
-                    .replace("...", "_catch_all_")
-                    .replace(']', "")
-                    .replace("__", "_"),
+                    .replace("...", "catch_all_")
+                    .replace(']', ""),
                 axum_route: axum_route
                     .replace("[...", "*")
                     .replace('[', ":")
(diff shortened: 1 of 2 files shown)


crates/tuono/src/route.rs
```diff
@@ -261,4 +262,12 @@ mod tests {
             assert_eq!(route.output_file_path(), PathBuf::from(html))
         }
     }
+
+    #[test]
+    fn should_correctly_replace_double_underscore() {
+        let dyn_info = AxumInfo::new(&Route::new("/[posts__id]".to_string()));
```
Valerioageno, line 268: What about also testing the scenario with `dyn_` + `_catch_all` that is more similar to the real bug?

Valerioageno, line 268: I see a problem here. The user can define both `post__id` and `post_id` route, and they resolve with the same name.

How about replacing the `__` with the `u_` string?

Valerioageno: I think the best solution would be to remove the `_` from the replacement prefix. `catch_all_`and `hyphen_`

mustafasegf: hmm yeah that sounds more reasonable. let me remove the underscore prefix on the replacement logic

mustafasegf: one small issue is the generated name will be a bit harder to read for hypen
`"/posts/handle-this/[post]" => "posts_handlehyphen_this_dyn_post"`
the case I see where double underscore will happen on hypen if it's use at thestartd of route name in dynamic route or catch all route. where catch all does need dyn for it to work. i think it's fine to only change the `_catch_all_` to `catch_all_`

Valerioageno: It sounds the best. Agree

Valerioageno (approved):
LGTM. Thanks

PR #233, opened 2024-12-18
feat: update rendering from `renderToString` to `renderToReadableStream`

Valerioageno:
## Context & Description

This PR turns the kinda "deprecated" `renderToString` to the new `renderToReadableStream` solving the last issue related to https://github.com/tuono-labs/tuono/issues/216.

For this iteration, we just wait for the whole stream, and then we pass it to the rust server. In a next iteration, we will need to pass every stream's chunk to rust to start streaming the response to the client. This will leverage the async server side resources loading, enhancing the user experience.

### DIFF
--- a/packages/tuono/src/build/index.ts
+++ b/packages/tuono/src/build/index.ts
@@ -1,6 +1,7 @@
-import type { InlineConfig } from 'vite'
+import type { InlineConfig, Plugin } from 'vite'
 import { build, createServer, mergeConfig } from 'vite'
 import react from '@vitejs/plugin-react-swc'
+import inject from '@rollup/plugin-inject'
 import ViteFsRouter from 'tuono-fs-router-vite-plugin'
 import { LazyLoadingPlugin } from 'tuono-lazy-fn-vite-plugin'
 
@@ -9,6 +10,14 @@ import type { TuonoConfig } from '../config'
 import { loadConfig, blockingAsync } from './utils'
 
 const VITE_PORT = 3001
+const VITE_SSR_PLUGINS: Array<Plugin> = [
+  {
+    enforce: 'post',
+    ...inject({
+      ReadableStream: ['web-streams-polyfill', 'ReadableStream'],
+    }),
+  },
+]
 
 /**
  * From a given {@link TuonoConfig} return a `vite` "mergeable" {@link InlineConfig}
@@ -65,6 +74,7 @@ const developmentSSRBundle = (): void => {
       mergeConfig<InlineConfig, InlineConfig>(
         createBaseViteConfigFromTuonoConfig(config),
         {
+          plugins: VITE_SSR_PLUGINS,
           build: {
             ssr: true,
             minify: false,
@@ -143,6 +153,7 @@ const buildProd = (): void => {
       mergeConfig<InlineConfig, InlineConfig>(
         createBaseViteConfigFromTuonoConfig(config),
         {
+          plugins: VITE_SSR_PLUGINS,
           build: {
             ssr: true,
             minify: true,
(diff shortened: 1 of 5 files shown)


packages/tuono/src/build/index.ts
```diff
@@ -1,6 +1,7 @@
-import type { InlineConfig } from 'vite'
+import type { InlineConfig, Plugin } from 'vite'
 import { build, createServer, mergeConfig } from 'vite'
 import react from '@vitejs/plugin-react-swc'
+import inject from '@rollup/plugin-inject'
 import ViteFsRouter from 'tuono-fs-router-vite-plugin'
 import { LazyLoadingPlugin } from 'tuono-lazy-fn-vite-plugin'
```
marcalexiei, line 6: Unrelated, but just noticed that `tuono-lazy-fn-vite-plugin` has a named export, 
while `tuono-fs-router-vite-plugin` has a default export. I think we should use named export in both.

I'll do it in a separate if you agree

packages/tuono/src/build/index.ts
```diff
@@ -65,6 +74,7 @@ const developmentSSRBundle = (): void => {
       mergeConfig<InlineConfig, InlineConfig>(
         createBaseViteConfigFromTuonoConfig(config),
         {
+          plugins: VITE_SSR_PLUGINS,
```
marcalexiei, line 77: Edge case: what if user specify an inject plugin with the same polypill?

packages/tuono/src/ssr/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+async function streamToArrayBuffer(
+  stream: ReadableStream<Uint8Array>,
+): Promise<Uint8Array> {
+  const chunks: Array<Uint8Array> = []
+
+  // @ts-expect-error ReadableStream has AsyncIterator https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration
+  // eslint-disable-next-line @typescript-eslint/await-thenable
+  for await (const chunk of stream) {
+    chunks.push(chunk as Uint8Array)
+  }
```
marcalexiei, line 20: Probably we should update `tsconfig` `compilerOptions#lib`  to include `AsyncIterator` types

packages/tuono/src/ssr/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * NOTE: we should improve the bond between the custom V8 runtime and the
+ * renderToReadableStream React function to return a stream directly to the client.
+ */
+export async function streamToString(
+  stream: ReadableStream<Uint8Array>,
+): Promise<string> {
+  const buffer = await streamToArrayBuffer(stream)
+  return new TextDecoder().decode(buffer)
+}
```
marcalexiei, line 36: I would consider removing `concatArrayBuffers` and `streamToArrayBuffer` and merge their code all in `streamToString` function.

Right now the functions are pretty simple so there is no need to split the code in 3 parts. WDYT?

packages/tuono/src/build/index.ts
Valerioageno: Agree. Thank you

packages/tuono/src/build/index.ts
Valerioageno: Well, why the user should ever do that? We basically provide that out of the box to them. I'd consider ignoring this for now and waiting to see how the users will behave.

packages/tuono/src/ssr/utils.ts
Valerioageno: Yeah. I left separate since it is more verbose this way. This code does very weird things, very uncommon. I think a bit more meaning could be helpful while reading this code in the future

marcalexiei: the final function has only two  rows so I would remove at least one, but I get your point we can leave this as it is!

packages/tuono/src/build/index.ts
marcalexiei: #234

marcalexiei (approved):

PR #237, opened 2024-12-19
chore: add CODEOWNERS

marcalexiei:
## Context & Description

To make PR handling easier I added a CODEOWNERS file.

[To have a complete overview about how this works you can visit the official documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file)

### Relevant references from the doc that helps understand how the files works:

```text
# These owners will be the default owners for everything in
# the repo. Unless a later match takes precedence,
# @global-owner1 and @global-owner2 will be requested for
# review when someone opens a pull request.
*       @global-owner1 @global-owner2

# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner #This is an inline comment.

# In this example, @doctocat owns any files in the build/logs
# directory at the root of the repository and any of its
# subdirectories.
/build/logs/ @doctocat
```

### Breakdown 

- If someone change **only** something inside `.github` I will be added as PR reviewer
- If someone change **only** something inside `crates` or the root `Cargo.tml` Valerioageno will be added as PR reviewer
- If other files are present in the PR will be both added as reviewers

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,9 @@
+*          @Valerioageno @marcalexiei
+
+# Rust
+/crates/    @Valerioageno
+/Cargo.toml @Valerioageno
+
+# Misc
+/.github/   @marcalexiei
+


Valerioageno (approved):
Sounds perfect. Thank you!

Issue #218, opened 2024-12-10
Add support React19


Valerioageno:
[React19](https://react.dev/blog/2024/12/05/react-19) just got release and this has to be fully supported by `tuono` out of the box. 

With this iteration, we are not expecting to support neither `server actions` nor `server components`. We just want to fully support server side rendering.

---

Before moving forward the followings are needed:
- [x] #250
- [x] #244
   - #254
- [x] #247
   - #256
- [ ] #264
  -  update `@testing-library/react`
  -  remove `react-async-helmet` that does not support React19
  -  update all the react version to the 19th

## Followup

- Update version to `v0.17.0`
- Update documentation

spacecodeur:
Given :  
- a new project created via `tuono new`
- The content of "package.json" file changed like this : 

```json
{
  "name": "tuono-app",
  "description": "Basic tuono application",
  "version": "0.0.1",
  "dependencies": {
    "@lit-labs/ssr-dom-shim": "^1.2.1",
    "react": "^19.0.0",
    "react-dom": "^19.0.0",
    "tuono": "0.16.0"
  },
  "devDependencies": {
    "@rollup/plugin-inject": "^5.0.5",
    "@types/react": "^19.0.1",
    "@types/react-dom": "^19.0.2",
    "typescript": "^5.7.2"
  }
}
```

When I run the app with `tuono dev`. No error occurs on the terminal
But when I load a new page on my browser to `http://localhost:3000`, the page doesn't load and this error occurs in my terminal : 

```
 <unknown>:269779: Uncaught ReferenceError: MessageChannel is not defined
thread 'tokio-runtime-worker' panicked at /home/okli/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tuono_lib-0.16.1/src/ssr.rs:54:10:
called `Result::unwrap()` on an `Err` value: InvalidJs("Execute your script with d8 to debug")
```

@Valerioageno on Discord : you suggested to follow [this way](https://github.com/tuono-labs/tuono/issues/204#issuecomment-2525615013) to fix this issue. 
I looked for a library that handle the internface MessageChannel in SSR context, but without success. 

I'll take a look at the Next.js repository that handles React version 19. Maybe I'll find something there about how they managed the MessageChannel interface. I can't promise anything, though—I’ve never developed this kind of thing before ;)

mustafasegf:
[...]

PR #238, opened 2024-12-19
feat: add MessageChannel polyfill to support react 19

mustafasegf:
add MessageChannel polyfill to support react 19
Related to #218

### DIFF
--- a/packages/tuono/package.json
+++ b/packages/tuono/package.json
@@ -106,6 +106,7 @@
     "@types/node": "^22.0.0",
     "@vitejs/plugin-react-swc": "^3.7.0",
     "fast-text-encoding": "^1.0.6",
+    "message-port-polyfill": "^0.2.0",
     "react-helmet-async": "^2.0.5",
     "tuono-fs-router-vite-plugin": "workspace:*",
     "tuono-lazy-fn-vite-plugin": "workspace:*",
--- a/packages/tuono/package.json
+++ b/packages/tuono/package.json
@@ -106,7 +106,6 @@
     "@types/node": "^22.0.0",
     "@vitejs/plugin-react-swc": "^3.7.0",
     "fast-text-encoding": "^1.0.6",
-    "message-port-polyfill": "^0.2.0",
     "react-helmet-async": "^2.0.5",
     "tuono-fs-router-vite-plugin": "workspace:*",
     "tuono-lazy-fn-vite-plugin": "workspace:*",
--- a/packages/tuono/src/build/index.ts
+++ b/packages/tuono/src/build/index.ts
@@ -15,6 +15,7 @@ const VITE_SSR_PLUGINS: Array<Plugin> = [
     enforce: 'post',
     ...inject({
       ReadableStream: ['web-streams-polyfill', 'ReadableStream'],
+      MessageChannel: ["message-port-polyfill", "MessageChannelPolyfill"],
     }),
   },
 ]
--- a/packages/tuono/src/build/index.ts
+++ b/packages/tuono/src/build/index.ts
@@ -15,7 +15,12 @@ const VITE_SSR_PLUGINS: Array<Plugin> = [
     enforce: 'post',
     ...inject({
       ReadableStream: ['web-streams-polyfill', 'ReadableStream'],
-      MessageChannel: ["message-port-polyfill", "MessageChannelPolyfill"],
+
+      /**
+       * Added to support `react@19`
+       * @see https://github.com/tuono-labs/tuono/issues/218
+       */
+      MessageChannel: ['message-port-polyfill', 'MessageChannelPolyfill'],
     }),
   },
 ]
--- a/packages/tuono/src/build/index.ts
+++ b/packages/tuono/src/build/index.ts
@@ -20,7 +20,7 @@ const VITE_SSR_PLUGINS: Array<Plugin> = [
        * Added to support `react@19`
        * @see https://github.com/tuono-labs/tuono/issues/218
        */
-      MessageChannel: ['message-port-polyfill', 'MessageChannelPolyfill'],
+      MessageChannel: ['tuono/ssr', 'MessageChannelPolyfill'],
     }),
   },
 ]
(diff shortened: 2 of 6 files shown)


packages/tuono/package.json
```diff
@@ -106,6 +106,7 @@
     "@types/node": "^22.0.0",
     "@vitejs/plugin-react-swc": "^3.7.0",
     "fast-text-encoding": "^1.0.6",
+    "message-port-polyfill": "^0.2.0",
```
Valerioageno, line 109: Is there any better library? Seems pretty unmaintained https://github.com/rocwind/message-port-polyfill

packages/tuono/src/build/index.ts
```diff
@@ -15,6 +15,7 @@ const VITE_SSR_PLUGINS: Array<Plugin> = [
     enforce: 'post',
     ...inject({
       ReadableStream: ['web-streams-polyfill', 'ReadableStream'],
+      MessageChannel: ["message-port-polyfill", "MessageChannelPolyfill"],
```
marcalexiei, line 18:
```suggestion

      /**
        * Added to support `react@19`
        * @see https://github.com/tuono-labs/tuono/issues/218
        */
      MessageChannel: ["message-port-polyfill", "MessageChannelPolyfill"],
```

marcalexiei (commented):
Looks good. Just a small comment.
Also, could you please apply correct formatting?

packages/tuono/package.json
marcalexiei: We could consider to creare our pollyfill since this is something we have to rely on to support react@19

marcalexiei: our pollyfill = fork that library (in here on in another repo) and use it here 😅

mustafasegf: yeah, it seems the library is simple enough that we can just fork it as a single file polyfill. we can create a `polyfills` folder in `packages/tuono/bin` for future polyfills files that we want to maintain ourselves

packages/tuono/src/build/index.ts
mustafasegf: ah my bad, i forgot to format again 😅

marcalexiei: no problem 😉

packages/tuono/package.json
marcalexiei: How about exporting the polyfill from `tuono/ssr`?

```diff
- MessageChannel: ['message-port-polyfill', 'MessageChannelPolyfill'], 
+ MessageChannel: ['tuono/ssr', 'MessageChannelPolyfill'],
```

I would add the polypill in a new `packages/tuono/src/ssr/polifylls` folder.
If you can add also some unit test it would be awesome!

@mustafasegf Would you like to try implementing this change?

mustafasegf: sure i can do that

marcalexiei:
@mustafasegf if you have finished, remember to request a new review from me and @Valerioageno!
I’d like to avoid reviewing code that’s still being worked on 😂

Also format (and probably lint, I see some import order that do not follow our custom order) check are failing. 
Could you please fix them?

mustafasegf:
> @mustafasegf if you have finished, remember to request a new review from me and @Valerioageno! I’d like to avoid reviewing code that’s still being worked on 😂
> 
> Also format (and probably lint, I see some import order that do not follow our custom order) check are failing. Could you please fix them?

haha yep, I'm still fixing the linting and typescript type error issue. I'll request for review after I'm done

Valerioageno:
I got this console error when trying your changes on `tuono-tutorial`. Do you have the same?

Screenshot 2024-12-20 203804

Valerioageno:
I checked `tuono-app` and `tuono-tutorial` by putting `react` and `react-dom` at `v19.0.0`

marcalexiei (commented):
Just submitting my review. You should just revert the changes on the `pnpm-lock` file since no dependencies are changed.

I also performed some debug on the issue Valerio reported but I'm posting what I discovered in a separate comment.

> [!WARNING]
> Note for myself: after closing the issue open a task to address review comment posted below

marcalexiei:
The error Valerio encountered is caused by changing `react@19` only on `examples/tuono-app` package.
However when printing the React version from `examples/tuono-app/src/routes/__root.tsx` the results is that the actual loaded `react` version is `18.3.1` not `19`:


(The object logs are the children printed inside RootRoute component)

image

### Why this happen?

<details>
<summary>
It happens due to the fact that examples reference `tuono` folder within the repository which contains his own version of `React`. I added a screenshot inside this `details` to show this.
</summary>

Screenshot 2024-12-21 at 14 37 56

</details>

To use the right version of `React` the right command is:

```sh
pnpm --filter="./examples/*" --filter="./packages/*" up react@latest react-dom@latest @types/react@latest @types/react-dom@latest
```

This should ensure that same version of React is used across all involved packages.

----

### Now it works right?

Unfortunately, no:
when running `tuono dev` from `examples/tuono-app`  the result is a React minified error printed in the HTML page:

image


```text
Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7B%24%24typeof%2C%20type%2C%20key%2C%20ref%2C%20props%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

[The links leads here](https://react.dev/errors/31?invariant=31&args%5B%5D=object%20with%20keys%20%7B%24%24typeof%2C%20type%2C%20key%2C%20ref%2C%20props%7D)

The error is the same we saw earlier:

> Objects are not valid as a React child (found: object with keys {$$typeof, type, key, ref, props}). If you meant to render a collection of children, use an array instead.

But now is printed directly in the HTML page.

---

### What to do next?

I think this PR can be safely merged (after `pnpm-lock.yaml` changes are reverted).
It should not have impact on `react@18`, which doesn't use `MessageChannel` API,
hence this polypill won't be included in the bundle.

Then we can figure out the next steps to support react@19.

WDYT?

Valerioageno:
Sounds perfect! Thanks for it.

pnpm-lock.yaml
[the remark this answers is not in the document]
mustafasegf: hmm, i just realize the lock filed got changed after I remove the `message-port-polyfill`. let me revert it

Valerioageno (approved):
LGTM. Thanks for it!

marcalexiei (approved):

PR #243, opened 2024-12-22
refactor: enable typescript-eslint  strict rules

marcalexiei:
## Context & Description

Followup of https://github.com/tuono-labs/tuono/pull/238#discussion_r1894311382

- enabled `strictTypeChecked` config rule
- remove redundant rules definition from `eslint.config.mjs`
- resolved new eslint errors coming from the following rules:
  - `@typescript-eslint/no-confusing-void-expression`
  - `@typescript-eslint/no-explicit-any`


packages/fs-router-vite-plugin/src/build-route-config.ts
```diff
@@ -3,7 +3,7 @@ import type { RouteNode } from './types'
 
 export function buildRouteConfig(nodes: Array<RouteNode>, depth = 1): string {
   const children = nodes.map((node) => {
-    const route = `${node.variableName}Route`
+    const route = `${node.variableName as string}Route`
```
Valerioageno, line 6: Why we need to cast the `string` here?

marcalexiei: https://github.com/tuono-labs/tuono/blob/de74647c434a0188951f7615b0e12f457a435186/packages/fs-router-vite-plugin/src/types.ts#L10

Because `variableName` is `string | undefined`

Valerioageno: Makes sense

Valerioageno (approved):
LGTM

Issue #247, opened 2024-12-22
[Feature request]: tuono-router - remove `zustand` and use React Context API


marcalexiei:
## Is your feature request related to a problem? Please describe.

Remove `zustand` and use React.context API instead

## Describe the solution you'd like

Right now `zustand` is used to keep track of the router state but all the logic can be moved inside a
new `InternalRouterContext`.

## Data of `RouterContext`

- router: it will be the router instance 
- state:
  https://github.com/tuono-labs/tuono/blob/77e22c40b6a0bb949e411e7e2ec26c70433c94f2/packages/router/src/hooks/useRouterStore.tsx#L14-L24
  All this properties will have their dedicated state (probably using `React.useState`)
  We might use a reducer but I think it might be add more overhead.

## Context interactions (read and update data)

> [!NOTE]
> After a search I noticed that only `location` and `updateLocation` are used

> [!NOTE] 
> The context itself will not be exported from the file. 
> To interact with this context we will have to go through specific hooks (listed below)

| Name | Current name | Description | Exported |
|--------|--------|--------|--------|
| `useRouterInstance` | useInternalRouter | provides the router instance |❌ |
| `useRouterLocation` | useRouter | provides the router location and the setter method and additional methods (`push`) | ✅ | 

----

With this refactor all router logic will be moved in a single place instead of be around different files:
- `packages/router/src/hooks/useRouterStore.tsx`
- `packages/router/src/hooks/useInternalRouter.tsx`
- `packages/router/src/components/RouterContext.tsx`
- `packages/router/src/components/RouterProvider.tsx`

## Additional context

<details>

<summary>Outdated considerations</summary>

`eslint-plugin-react-hooks` is not present in the project and I found some violations about the rules of hooks:
1. `useMemo` is conditional here: https://github.com/tuono-labs/tuono/blob/77e22c40b6a0bb949e411e7e2ec26c70433c94f2/packages/router/src/components/RouteMatch.tsx#L60-L68
2. This layout effect should have `updateLocation`, the function should be stable so it should not change hence it can be safely added to the dependencies https://github.com/tuono-labs/tuono/blob/77e22c40b6a0bb949e411e7e2ec26c70433c94f2/packages/router/src/hooks/useRouterStore.tsx#L43-L52

The code is probably working in the current state but, without rule of Hooks static check and the absence of unit tests, performing this kind of refactor looks very error prone to me.

</details>

Resolved by:
- #248
- #251

----

## What to do next?

[...]

PR #248, opened 2024-12-22
refactor: add `eslint-plugin-react-hooks`

marcalexiei:
## Context & Description

Related to #247.

- Add `eslint-plugin-react-hooks`
- Remove `eslint-plugin-react` since it is not used


packages/router/src/components/RouteMatch.tsx
```diff
@@ -58,13 +53,7 @@ const TraverseRootComponents = React.memo(
...
   }: TraverseRootComponentsProps): React.JSX.Element => {
     if (routes.length > index) {
-      const Parent = React.useMemo(
-        () =>
-          routes[index]?.component as unknown as (
-            props: ParentProps,
-          ) => React.JSX.Element,
-        [],
-      )
+      const Parent = (routes[index] as Route).component
```
marcalexiei, line 56: - The optional chain should not be needed: if route is `undefined` the `<Parent>` below should have throw an error since `undefined` is not a `React.ElementType`
- The `useMemo` has no dependencies, so if the `routes` or `index` prop changes the memo will still have the inital reference which might cause some bug or gc issues

Due to the previous points removing `useMemo` seems the most viable way to resolve the rule of hooks error

Valerioageno:
Changes QAed ✅

Valerioageno (approved):
PR looks good! Thanks for it

PR #250, opened 2024-12-22
Add message event polyfill

Valerioageno:
## Context & Description

This PR does:
- Update the `tuono/ssr` folder structure
- Implements the `Event` and `MessageEvent` interfaces polyfills that will be needed for React19 SSR

The polyfills have been inspired by [this package](https://www.npmjs.com/package/event-source-polyfill), but I ended up creating our own classes that are more typescript complaint


packages/tuono/src/ssr/index.ts
```diff
@@ -0,0 +1,4 @@
+export { MessageChannelPolyfill } from './polyfills/MessageChannel'
+export { EventPolyfill } from './polyfills/Event'
+export { MessageEventPolyfill } from './polyfills/MessageEvent'
+export { serverSideRendering } from './server'
```
marcalexiei, line 4:
```suggestion
export { serverSideRendering } from './server'

export { MessageChannelPolyfill } from './polyfills/MessageChannel'
export { EventPolyfill } from './polyfills/Event'
export { MessageEventPolyfill } from './polyfills/MessageEvent'
```

Style: just reorder import adding groups and putting `serverSideRendering` as first

packages/tuono/src/build/index.ts
```diff
@@ -21,6 +21,8 @@ const VITE_SSR_PLUGINS: Array<Plugin> = [
        * @see https://github.com/tuono-labs/tuono/issues/218
        */
       MessageChannel: ['tuono/ssr', 'MessageChannelPolyfill'],
+      MessageEvent: ['tuono/ssr', 'MessageEventPolyfill'],
+      Event: ['tuono/ssr', 'EventPolyfill'],
```
marcalexiei, line 25: Do we need to create a task to port these polypill in rust in the future?

marcalexiei (commented):
Just a minor style change and a task related question. 
Everything else looks good!

packages/tuono/src/ssr/index.ts
Valerioageno: What about adding a formatting rule for this?

packages/tuono/src/build/index.ts
Valerioageno: I don't know but I would not think so because they just hold values...
I think it could be worth running an investigation in the future but not sure if addressing them makes sense. WDYT?

packages/tuono/src/ssr/index.ts
marcalexiei: I'm not aware of any way to statically define and organise exports via eslint / prettier. 
Those kind of rules are available for import, not for exports

packages/tuono/src/build/index.ts
marcalexiei: Ok, no need to track this for now. We already too much work ahead of us 😂

packages/tuono/src/ssr/index.ts
Valerioageno: How about this one? 

https://www.npmjs.com/package/eslint-plugin-sort-exports

marcalexiei: I have seen it, but I haven't proposed it because:
- looks un-maintened (no commit in nearly 1 year) 
- examples refer only to legacy config and not to flat config

If you want I can give it a try anyway 
(since it exports only one rule there should be no problem using in with the flat config)

Valerioageno: Yeah I think it would be worth giving it a try. 
In the end it is just matter of reorder the exports. 
In the download chart sounds pretty used still

marcalexiei: Ok, I'll take a look at it tomorrow, meanwhile I'll go ahead and manually apply the fix so we can merge this.

marcalexiei (approved):

PR #256, opened 2024-12-23
feat(packages/tuono-router): replace `zustand` with `React.Context`

Valerioageno:
## Context & Description

As required by the React19 migration, this PR turn the router zustand store into a react context.

Closes: https://github.com/tuono-labs/tuono/issues/247


packages/tuono-router/package.json
```diff
@@ -48,8 +48,7 @@
   },
   "dependencies": {
     "react-intersection-observer": "^9.13.0",
-    "vite": "^5.2.11",
-    "zustand": "4.4.7"
+    "vite": "^5.2.11"
```
Valerioageno, line 51: I think this can become a `devDependencies`

marcalexiei: Yes, also set it to a fixed version range: right now we have `5.4.11` installed

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -1,13 +1,29 @@
...
+
+export interface RouterContextType {
+  router: Router
+  location: ParsedLocation
+  updateLocation: (loc: ParsedLocation) => void
+}
 
 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-const routerContext = React.createContext<Router>(null!)
+const routerContext = React.createContext<RouterContextType>(null!)
```
marcalexiei, line 22: [Context should have a PascalCase name](https://react.dev/reference/react/createContext)

```suggestion
const RouterContext = React.createContext<RouterContextType>(null!)
```

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -1,13 +1,29 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
```
marcalexiei, line 1: Style: `useState` and `useEffect` are imported via named import while all context functions are prefixed with `React.`. 
I think we should be more consistent. 
I usually prefer importing `React` and prefix all with `React.` (also properties which starts with `React`. It's redundant, but at least the style is consistent on all files).
There are types like `ElementType` or handlers (like `MouseEventHandler`, ...) that cannot be associated with React when used without `React.`

(I can take care of this in a separate PR)

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -1,13 +1,29 @@
...
+
+export interface ParsedLocation {
+  href: string
+  pathname: string
+  search: Record<string, string>
+  searchStr: string
+  hash: string
+}
+
+export interface RouterContextType {
```
marcalexiei, line 15:
```suggestion
export interface RouterContextValue {
```

I think we should changes this:
`RouterContextType` is an interface not a type.
I'm not a big fan of the [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) like variables name

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -20,3 +36,88 @@ export function getRouterContext(): React.Context<Router> {
...
+
+export function RouterContextProvider({
+  router,
+  children,
+  serverSideProps,
+}: RouterContextProviderProps): ReactNode {
+  // Allow the router to update options on the router instance
+  router.update({ ...router.options } as Parameters<typeof router.update>[0])
+
+  const [location, updateLocation] = useState<ParsedLocation>(
```
marcalexiei, line 78:
```suggestion
  const [location, setLocation] = useState<ParsedLocation>(
```

This can be renamed when creating the `RouterContext` value to useLocation, but let's stick to react conventions when declaring state variables

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -20,3 +36,88 @@ export function getRouterContext(): React.Context<Router> {
 
   return routerContext
 }
+
+interface RouterContextProviderProps {
+  router: Router
+  children: ReactNode
+  serverSideProps?: ServerRouterInfo
+}
```
marcalexiei, line 44: This should be moved right before `RouterContextProvider`. 
Props should be alongside component declarations.
```suggestion
```

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -20,3 +36,88 @@ export function getRouterContext(): React.Context<Router> {
...
+  const { location } = window
+  return {
+    pathname: location.pathname,
+    hash: location.hash,
+    href: location.href,
+    searchStr: location.search,
+    search: Object.fromEntries(new URLSearchParams(location.search)),
+  }
+}
+
```
marcalexiei, line 69:
```suggestion

interface RouterContextProviderProps {
  router: Router
  children: ReactNode
  serverSideProps?: ServerRouterInfo
}

```

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -20,3 +36,88 @@ export function getRouterContext(): React.Context<Router> {
...
+  const RouterContext = getRouterContext()
+
+  const context: RouterContextType = {
+    router,
+    location,
+    updateLocation,
+  }
+
+  return (
+    <RouterContext.Provider value={context}>{children}</RouterContext.Provider>
```
marcalexiei, line 117:
```suggestion
  const contextValue: RouterContextValue = {
    router,
    location,
    updateLocation,
  }

  return (
    <RouterContext.Provider value={contextValue}>{children}</RouterContext.Provider>
```

packages/tuono-router/src/components/RouterProvider.tsx
```diff
@@ -1,51 +1,32 @@
...
-  )
-}
-
 interface RouterProviderProps {
   router: Router
   serverProps?: ServerProps
 }
 
+/**
+ * This component is the tuono app entry point
```
marcalexiei, line 16: Just a small reword, otherwise it look like that we are targeting this file from the tuono build
```suggestion
 * This component is used in the tuono app entry point
```

packages/tuono-router/src/components/RouterProvider.tsx
```diff
@@ -1,51 +1,32 @@
...
 }: RouterProviderProps): JSX.Element {
-  useInitRouterStore(serverProps)
-
-  useListenBrowserUrlUpdates()
-
   return (
-    <RouterContextProvider router={router}>
-      <Matches serverSideProps={serverProps?.props} />
-    </RouterContextProvider>
+    <Suspense>
```
marcalexiei, line 23: IMHO the Suspense should not be added here. 
It's likely that a user wants to customise the `fallback` behaviour,
so I think is better if we handle this directly in `tuono` package

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -20,3 +36,88 @@ export function getRouterContext(): React.Context<Router> {
...
+  )
+}
+
+export function useInternalRouter(): RouterContextType {
```
marcalexiei, line 121: All names in this file speak of `RouterContext`. The word Internal comes out only here.
Since his hook is not exposed to the user I think we should named it simply `useRouterContext`.
To avoid accidental exports we might add a `@warning` comment

```suggestion
/** @warning DO NOT EXPORT THIS TO USER LAND */
export function useRouterContext(): RouterContextValue {
```

packages/tuono-router/src/hooks/useRoute.tsx
```diff
@@ -25,7 +25,8 @@ export function sanitizePathname(pathname: string): string {
  * Optimizations should occour on both
  */
 export default function useRoute(pathname?: string): Route | undefined {
-  const { routesById } = useInternalRouter()
+  const { router } = useInternalRouter()
+  const { routesById } = router
```
marcalexiei, line 29:
```suggestion
  const { router: { routesById } } = useInternalRouter()
```

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -1,13 +1,29 @@
...
+}
 
 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-const routerContext = React.createContext<Router>(null!)
+const routerContext = React.createContext<RouterContextType>(null!)
 
 const TUONO_CONTEXT_GLOBAL_NAME = '__TUONO_CONTEXT__'
 
-export function getRouterContext(): React.Context<Router> {
+function getRouterContext(): React.Context<RouterContextType> {
```
marcalexiei, line 26: I'm not sure about setting the whole context value in a window variable. 
How about adding only the router instance?

marcalexiei, line 26: How about removing this function and set window variable inside an `useEffect` in `RouterContextProvider`?
effect should be executed server side so this avoids the necessity to perform check on document / window existence

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -20,3 +36,88 @@ export function getRouterContext(): React.Context<Router> {
...
+      window.removeEventListener('popstate', updateLocationOnPopStateChange)
+    }
+  }, [updateLocation])
+
+  const RouterContext = getRouterContext()
+
+  const context: RouterContextType = {
+    router,
+    location,
+    updateLocation,
```
marcalexiei, line 113: Since this context wraps the whole application I would consider to wrap the context value in a `React.useMemo`

packages/tuono-router/src/hooks/useRouter.tsx
```diff
@@ -25,10 +25,7 @@ interface UseRouterHook {
 }
 
 export const useRouter = (): UseRouterHook => {
-  const [location, updateLocation] = useRouterStore((st) => [
-    st.location,
-    st.updateLocation,
-  ])
+  const { location, updateLocation } = useInternalRouter()
 
   const push = (path: string, opt?: PushOptions): void => {
```
marcalexiei, line 30: push should be a stable function. Right now is a new function on every render.
It should be wrapped in a `React.useCallback`

marcalexiei (changes requested):
Left a few point of discussions 😅.
If you agree with them I can take care of applying them to the PR.

packages/tuono-router/src/components/RouterProvider.tsx
Valerioageno: This Suspense is required by react. Without it we get a console error.

packages/tuono-router/src/components/RouterContext.tsx
Valerioageno: I was thinking to completely remove it. I think it was interesting for tanstack-router for their devtools but we don't need that.

packages/tuono-router/src/components/RouterContext.tsx
Valerioageno: What props might cause re-rendering?

packages/tuono-router/src/hooks/useRouter.tsx
Valerioageno: Agree

packages/tuono-router/src/components/RouterProvider.tsx
marcalexiei: Gotcha, we might end up to add a prop to allow `fallback` customisation in the future.

packages/tuono-router/src/components/RouterContext.tsx
marcalexiei: probably `children`, also the component is not wrapped in a `React.memo` so any change in the components up in the tree might cause its re-rendering.
I understand this is used in the app entry point and has a limited impact, but better be safe than sorry 😅

Valerioageno: Definitely!

packages/tuono-router/src/components/RouterContext.tsx
Valerioageno: I'd always use named imports.

packages/tuono-router/src/components/RouterContext.tsx
```diff
@@ -1,22 +1,111 @@
...
+): ParsedLocation {
   if (typeof document === 'undefined') {
-    return routerContext
+    return {
+      pathname: serverSideProps?.pathname || '',
+      hash: '',
+      href: serverSideProps?.href || '',
+      searchStr: serverSideProps?.searchStr || '',
+      // TODO: Polyfill URLSearchParams
+      search: {},
```
marcalexiei, line 34: Do we need to create a separate task for this?

marcalexiei (approved):
~~Left a comment regarding `URLSearchParams` polyfill.~~
Everything else looks good to me.

packages/tuono-router/src/components/RouterContext.tsx
marcalexiei: - #259

PR #260, opened 2024-12-24
feat(packages/vite-plugins): use named export and consistent names across plugins

### LINKED ISSUE #234: feat(packages/tuono-fs-router-vite-plugin): expose the plugin using a named export
but just noticed that `tuono-lazy-fn-vite-plugin` has a named export, 
while `tuono-fs-router-vite-plugin` has a default export. I think we should use named export in both.

I'll do it in a separate if you agree

_Originally posted by @marcalexiei in https://github.com/tuono-labs/tuono/pull/233#discussion_r1890667946_

marcalexiei:
## Context & Description

Close #234.

All vite plugins now have consistent names and they are made available via a named export.

### DIFF
--- a/packages/tuono-fs-router-vite-plugin/src/index.ts
+++ b/packages/tuono-fs-router-vite-plugin/src/index.ts
@@ -8,7 +8,7 @@ const ROUTES_DIRECTORY_PATH = './src/routes'
 
 let lock = false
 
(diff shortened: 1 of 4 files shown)


Valerioageno (approved):
Noice! Thanks

PR #264, opened 2024-12-24
feat: add support for react 19

marcalexiei:
# ✨🎉⚡️ Proof of concept with React 19 (it's working) ✨🎉⚡️

## Context & Description

This PR brings complete support to react 19 to `tuono`

---

> [!WARNING]
> We are dropping react < 19 support.
> Support both version is not worth at the current state in current state of the project
> We made a poll on Discord if you 

--- 

### Pending tasks

- [ ] Check that examples are working
   - [x] `tuono-tutorial`
   - [ ] `tuono-app`
   - [ ] `tuono-mdx`


packages/tuono/src/ssr/server.tsx
```diff
@@ -47,49 +40,21 @@ export function serverSideRendering(routeTree: RouteTree) {
     >
 
     const mode = serverProps.mode as Mode
-    const jsBundles = serverProps.jsBundles as Array<string>
-    const cssBundles = serverProps.cssBundles as Array<string>
```
marcalexiei, line 51: Since those two properties are not needed, 
I think we can remove the Rust code that is creating them

Valerioageno: They are needed!!!

During dev they are just ignored. On prod they Server side render the bundles of the requested page.

> My branch wasn't ready at all...

marcalexiei: Added a new task in the PR

marcalexiei:
## Avoid missing export error when running tuono dev

This error was caused by an awkward resolution when running `tuono dev` inside `examples/tuono-tutorial` folder: 

resolution error

As you can see from this image the error is coming from `tuono@0.16.4` which is the version used by the documentation...
this should not happen since `examples/tuono-tutorial` is linked to the `tuono` workspace within the repository.

To solve the issues:
1. In `apps/documentation` change `tuono` version to `workspace:*`
2. Run `pnpm I`
3. Revert `#1`
4. Run `pnpm I` again
5. Run `tuono dev` again. The issue is no longer present.

packages/tuono/src/ssr/server.tsx
```diff
@@ -1,9 +1,30 @@
...
+  scope: Partial<Pick<typeof globalThis, 'MessageChannel'>> = {},
+): void {
+  scope['MessageChannel'] = scope['MessageChannel'] ?? MessageChannelPolyfill
+})(
+  typeof window !== 'undefined'
+    ? window
+    : typeof global !== 'undefined'
+      ? global
+      : this,
+)
```
marcalexiei, line 22: I added an explanation to this workaround to provide `MessageChannel` polyfill.
Since it is a little bit "hacky" I would really consider to log a task to find a better alternative.

marcalexiei: #266

PR #265, opened 2024-12-25
chore: refine `eslint` ignores and fix format script of `tuono-fs-router-vite-plugin`

marcalexiei:
## Context & Description

Avoid to run `eslint` on the following folders:
- `out` - `tuono build` output folder
- `examples/*` workspaces, there is no script, not CI task that checks linting for those folder.
   We might want to change this in the future.

---

Fix format related scripts on `tuono-fs-router-vite-plugin`

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,4 +3,6 @@ pnpm-lock.yaml
 dist
 .tuono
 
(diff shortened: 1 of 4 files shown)


Valerioageno (approved):
LGTM. Thanks

PR #276, opened 2025-01-01
chore: dockerization of the tuono repo!

spacecodeur:
## Context & Description

I have dockerized the repo to simplify local development on the source code (one environment to rule them all!). To determine which packages to install and their order, I relied on the ['local development' documentation](https://tuono.dev/documentation/contributing/local-development).

Everything has been isolated in a dedicated docker folder (to avoid adding extra configuration files at the root of the project).

Currently, packages such as `pnpm`, `turbo`, and `cargo` are properly installed, and tuono (seems) compiles correctly. I noticed that the Node version is currently managed via the `/.nvmrc` file. To streamline the build process (and centralize versioning) with Docker, I have transferred this version into the `docker/compose.yml` file.

## How to use Docker (with this current work)

Firstly, install docker and docker compose. 
Then navigate to the root of the project using your terminal, then...

- Build the Docker container : `docker compose -f docker/compose.yml up --build -d`
- List active images and containers : `docker images && docker ps -a --size`
- Connect to the container : `docker exec -it tuono-source-container /bin/bash`
- (bonus) Delete all Docker images/containers/volumes + caches (caution here!) : 

```bash
docker ps -q | \
xargs -r docker stop && docker ps -aq | \
xargs -r docker rm -v && docker network prune -f && docker system prune -af --volumes && docker images -q | \
xargs -r docker rmi
```

The docker container's responsability is to handle all the project environment. That's mean you have only to just install your prefered IDE (like vscode by example), connect it to the docker container and the dev can begin ! The code source is sync between container and host ; at the end, you can git add/commit/push/... from your host machine.

## To do 

- I can't run the web server for the documentation website  (the `tuono dev` command seems work well, but a `curl http://127.0.0.1:3000` from container failed). I haven't found how to fix this problem yet..
- The command `turbo check-all` does not work (turbo error : could not find task check-all in project). I have to run directly `turbo build lint format:check types --filter='!./examples'`
- Since I don't know the Turbo tool very well, I haven't delved too deeply into it with Docker yet. I assume there's some work to be done on that front

Once the work with Docker is complete, I can suggest updating the documentation to summarize how to use it, if you'd like :)

### DIFF
--- a/apps/documentation/src/routes/documentation/contributing/local-development.mdx
+++ b/apps/documentation/src/routes/documentation/contributing/local-development.mdx
@@ -30,9 +30,9 @@ git clone https://github.com/<your-name-here>/tuono
 cd tuono
 ```
 
-For the next steps, you can either set up the environment directly on your computer or use a ready-to-use Docker image (go to [Use Docker](#use-docker) part).
+For the next steps, you can either set up the environment directly on your computer or use a ready-to-use Docker image (go to the [Docker setup](#docker-setup) part).
 
-### Set up your computer directly
+### Local setup
 
 #### Rust tool chain
 
@@ -69,7 +69,7 @@ pnpm run check-all
 cargo build
 ```
 
-### Use Docker
+### Docker setup
 
 #### Introduction
 
@@ -79,14 +79,14 @@ Some IDEs (such as vscode with the [Dev Containers extension](https://marketplac
 
 **Before proceeding with this guide, make sure that the `node_modules` and `target` directories are either absent or empty in the subprojects of Tuono from your host machine where you want to contribute.** This ensures a clean and consistent environment when using Docker.
 
-#### Build the Docker Container for Tuono
+#### Build the Tuono's Docker Container
 
 First, ensure Docker is installed on your machine by following the instructions on the [Docker official site](https://docs.docker.com/engine/install/).
 
 Once Docker is installed, use the following command to build the image into a container named tuono-source-container:
 
 ```sh
-# Note: Execute this command at the project root
+# Note: Execute this command in the project root
 docker compose -f docker/compose.yml up --build -d
 ```
 
@@ -96,13 +96,13 @@ To verify that everything is working as expected, run the following command:
 docker images && docker ps -a --size
 ```
 
-Or on Windows :
+Or on Windows:
 
 ```powershell
 docker images; docker ps -a --size
 ```
 
-You should see the image named tuono-source-image and the container tuono-source-container. The container's status should be "Up".
+You should see the image named tuono-source-image and the container tuono-source-container. The container's status should be "up".
 
 Use the following command to connect to the container:
 
--- a/apps/documentation/src/routes/documentation/contributing/local-development.mdx
+++ b/apps/documentation/src/routes/documentation/contributing/local-development.mdx
@@ -73,9 +73,9 @@ cargo build
 
 #### Introduction
 
-Docker takes care of configuring the development environment, so you don’t need to worry about installing specific versions of packages like Node.js or pnpm. The Docker image handles this for you! 
+Docker takes care of configuring the development environment, so you don’t need to worry about installing specific versions of packages like Node.js or pnpm. The Docker image handles this for you!
 Using Docker, **the only operations you’ll need to manage on your host machine are related to git**. (don't use phpnm, cargo, ... from your host machine)
-Some IDEs (such as vscode with the [Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)) allow you to connect directly to the Docker container. 
+Some IDEs (such as vscode with the [Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)) allow you to connect directly to the Docker container.
 
 **Before proceeding with this guide, make sure that the `node_modules` and `target` directories are either absent or empty in the subprojects of Tuono from your host machine where you want to contribute.** This ensures a clean and consistent environment when using Docker.
 
--- a/apps/documentation/src/routes/documentation/contributing/local-development.mdx
+++ b/apps/documentation/src/routes/documentation/contributing/local-development.mdx
@@ -30,12 +30,16 @@ git clone https://github.com/<your-name-here>/tuono
 cd tuono
 ```
 
-### Rust tool chain
+For the next steps, you can either set up the environment directly on your computer or use a ready-to-use Docker image (go to [Use Docker](#use-docker) part).
+
+### Set up your computer directly
+
+#### Rust tool chain
 
 Install the Rust programming language tool chain (`rust` and `cargo`).
 Follows instructions in the official [docs](https://rustup.rs/)
 
-### Node.js — runtime
+#### Node.js — runtime
 
 Install `Node.js`.
 You can follow the instructions from the [Node official site](https://nodejs.org/en/download/package-manager)
@@ -50,19 +54,58 @@ You can follow the instructions from the [Node official site](https://nodejs.org
 >
 > to simply pick up the correct version!
 
-### Node.js — package manager
+#### Node.js — package manager
 
 We use [`pnpm`](https://pnpm.io) as Node.js package manager.
 
 You can see which version of yarn we use by checking the `packageManager` field in the root `package.json`.
 
-### Pre-flight checks
+#### Pre-flight checks
 
 To check that everything is working properly, run:
 
 ```sh
 pnpm run check-all
 cargo build
+```
+
+### Use Docker
+
+#### Introduction
+
+Docker takes care of configuring the development environment, so you don’t need to worry about installing specific versions of packages like Node.js or pnpm. The Docker image handles this for you!
+Using Docker, the only operations you’ll need to manage on your host machine are related to git.
+Some IDEs (such as vscode with the [Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)) allow you to connect directly to the Docker container.
+
+#### Build the Docker Container for Tuono
+
+First, ensure Docker is installed on your machine by following the instructions on the [Docker official site](https://docs.docker.com/engine/install/).
+
+Once Docker is installed, use the following command to build the image into a container named tuono-source-container:
+
+```sh
+# Note: Execute this command at the project root
+docker compose -f docker/compose.yml up --build -d
+```
+
+To verify that everything is working as expected, run the following command:
+
+```sh
+docker images && docker ps -a --size
+```
+
+Or on Windows :
+
+```powershell
+docker images; docker ps -a --size
+```
+
+You should see the image named tuono-source-image and the container tuono-source-container. The container's status should be "Up".
+
+Use the following command to connect to the container:
+
+```sh
+docker exec -it tuono-source-container /bin/bash
 ```
 
 ## Tuono development
--- a/apps/documentation/src/routes/documentation/contributing/local-development.mdx
+++ b/apps/documentation/src/routes/documentation/contributing/local-development.mdx
@@ -73,9 +73,11 @@ cargo build
 
 #### Introduction
 
-Docker takes care of configuring the development environment, so you don’t need to worry about installing specific versions of packages like Node.js or pnpm. The Docker image handles this for you!
-Using Docker, the only operations you’ll need to manage on your host machine are related to git.
-Some IDEs (such as vscode with the [Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)) allow you to connect directly to the Docker container.
+Docker takes care of configuring the development environment, so you don’t need to worry about installing specific versions of packages like Node.js or pnpm. The Docker image handles this for you! 
+Using Docker, **the only operations you’ll need to manage on your host machine are related to git**. (don't use phpnm, cargo, ... from your host machine)
+Some IDEs (such as vscode with the [Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)) allow you to connect directly to the Docker container. 
+
+**Before proceeding with this guide, make sure that the `node_modules` and `target` directories are either absent or empty in the subprojects of Tuono from your host machine where you want to contribute.** This ensures a clean and consistent environment when using Docker.
 
 #### Build the Docker Container for Tuono
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -46,5 +46,5 @@ RUN ["pnpm", "install", "--frozen-lockfile"]
 
 ## Compile and add the directory with compiled binaries to the PATH for direct access
 
-RUN ["cargo", "build", "--frozen-lockfile"]
+RUN ["cargo", "build"]
 ENV PATH=/tuono/target/debug:$PATH
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,42 +1,50 @@
-# Use the specified Rust version and base OS as the base image
+# ARGs obtained from the compose.yml file
+
 ARG BASE_OS
 ARG RUST_VERSION
+
 FROM rust:${RUST_VERSION}-${BASE_OS}
 
 # Nodejs installation
-RUN ["mkdir", "/usr/local/nvm"]
-ENV NVM_DIR /usr/local/nvm
 
-RUN ["apt", "update"]
-RUN ["apt", "install", "curl", "-y", "--no-install-recommends"]
+RUN ["mkdir", "/usr/local/node"]
+WORKDIR /usr/local/node
 
-RUN ["curl", "https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh", "-o", "/tmp/install_nvm.sh"]
-RUN ["bash", "/tmp/install_nvm.sh"]
+## Retrieves the targeted version of Node in the container
 
-ARG NODE_VERSION
-RUN ["sh", "-c", ". /usr/local/nvm/nvm.sh && nvm install ${NODE_VERSION}"]
+COPY .nvmrc /tmp/.nvmrc
 
-# Ensure Node.js binaries are available in the PATH
-ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
+## Download the Node.js standelone binary archive && extract it && clean files
 
-# Install pnpm
+RUN ["/bin/sh", "-c", "NODE_VERSION=$(cat /tmp/.nvmrc) && \
+    wget https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz && \
+    tar xf node-v${NODE_VERSION}-linux-x64.tar.xz --strip-components 1 && \
+    rm node-v${NODE_VERSION}-linux-x64.tar.xz /tmp/.nvmrc"]
+
+## Ensure Node.js binaries are available in the PATH
+
+ENV PATH /usr/local/node/bin:$PATH
+
+# pnpm installation
+
 RUN ["npm", "install", "-g", "pnpm"]
-RUN ["mkdir", "/pnpm"]
-ENV PATH=/pnpm:$PATH
-ENV PNPM_HOME=/pnpm
 
-# Install cargo-watch for Rust live-reloading
+# cargo-watch installation (for Rust live-reloading during dev')
+
 RUN ["cargo", "install", "cargo-watch"]
 
-# Set the working directory to /tuono
-WORKDIR /tuono
+# global configuration done ! let's focus on the project itself
 
-# Copy project files into the container (excepted files/dir declared in .dockerignore)
+## Copy project files into the container (excepted files/dir declared in .dockerignore) to /tuono directory
+
+WORKDIR /tuono
 COPY . .
 
-# Install Node.js dependencies (defined in package.json)
+## Install Node.js dependencies (defined in package.json)
+
 RUN ["pnpm", "install"]
 
-# Compile and add the directory with compiled binaries to the PATH for direct access
+## Compile and add the directory with compiled binaries to the PATH for direct access
+
 RUN ["cargo", "build"]
 ENV PATH=/tuono/target/debug:$PATH
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,12 +19,11 @@ RUN ["sh", "-c", ". /usr/local/nvm/nvm.sh && nvm install ${NODE_VERSION}"]
 # Ensure Node.js binaries are available in the PATH
 ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
 
-# Install pnpm and turbo globally
+# Install pnpm
 RUN ["npm", "install", "-g", "pnpm"]
 RUN ["mkdir", "/pnpm"]
 ENV PATH=/pnpm:$PATH
 ENV PNPM_HOME=/pnpm
-RUN ["pnpm", "install", "-g", "turbo"]
 
 # Install cargo-watch for Rust live-reloading
 RUN ["cargo", "install", "cargo-watch"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,10 +25,6 @@ RUN ["/bin/sh", "-c", "NODE_VERSION=$(cat /tmp/.nvmrc) && \
 
 ENV PATH /usr/local/node/bin:$PATH
 
-# pnpm installation
-
-RUN ["npm", "install", "-g", "pnpm"]
-
 # cargo-watch installation (for Rust live-reloading during dev')
 
 RUN ["cargo", "install", "cargo-watch"]
@@ -39,6 +35,11 @@ RUN ["cargo", "install", "cargo-watch"]
 
 WORKDIR /tuono
 COPY . .
+
+# pnpm installation based on version found in package.json
+
+RUN ["corepack", "enable"]
+RUN ["corepack", "install"]
 
 ## Install Node.js dependencies (defined in package.json)
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,43 @@
+# Use the specified Rust version and base OS as the base image
+ARG BASE_OS
+ARG RUST_VERSION
+FROM rust:${RUST_VERSION}-${BASE_OS}
+
+# Nodejs installation
+RUN ["mkdir", "/usr/local/nvm"]
+ENV NVM_DIR /usr/local/nvm
+
+RUN ["apt", "update"]
+RUN ["apt", "install", "curl", "-y", "--no-install-recommends"]
+
+RUN ["curl", "https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh", "-o", "/tmp/install_nvm.sh"]
+RUN ["bash", "/tmp/install_nvm.sh"]
+
+ARG NODE_VERSION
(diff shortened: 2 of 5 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN ["mkdir", "/pnpm"]
+ENV PATH=/pnpm:$PATH
+ENV PNPM_HOME=/pnpm
+RUN ["pnpm", "install", "-g", "turbo"]
+
+# Install cargo-watch for Rust live-reloading
+RUN ["cargo", "install", "cargo-watch"]
+
+# Set the working directory to /app
+WORKDIR /app
```
marcalexiei, line 33: I would consider to change the `app` name.
We already have `apps` inside repository, so, IMHO it is a little bit confusing.

I would consider to use `tuono`  or `tuono-src`. WDYT?

marcalexiei (commented):
Thank you for this PR.

I left a few comments.

----

> The command turbo check-all does not work (turbo error : could not find task check-all in project). I have to run directly turbo build lint format:check types --filter='!./examples'

I just realised that the doc has an error.
`check-all` is not a turbo task. 
As you can see it is not listed in `turbo.json#tasks`:

https://github.com/tuono-labs/tuono/blob/f6593097b618e4c5dfff4d24671e09e7c96a0cc5/turbo.json#L4-L22

It should be executed via `pnpm` `run` command:
```sh
pnpm run check-all
```

I need to correct the doc. Thank you very much for bringing this up!

---

> I can't run the web server for the documentation website (the tuono dev command seems work well, but a curl http://127.0.0.1:3000 from container failed). I haven't found how to fix this problem yet..

Is the error something like this?

```text
thread 'main' panicked at crates/tuono/src/cli.rs:111:18:
Failed to build tuono.config.ts: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

It looks like that `pnpm install`  has not been executed in the repository:
there is no `node_modules` inside `apps/documentation`.

After running `pnpm install` manually in the `/app` folder and then `tuono dev` (from `apps/documentation`) I was able to successfully view the documentation site.

docker/Dockerfile
spacecodeur: agree with you
I'll replace `/app`by `tuono`

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# Ensure Node.js binaries are available in the PATH
+ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
+
+# Install pnpm and turbo globally
+RUN ["npm", "install", "-g", "pnpm"]
+RUN ["mkdir", "/pnpm"]
+ENV PATH=/pnpm:$PATH
+ENV PNPM_HOME=/pnpm
+RUN ["pnpm", "install", "-g", "turbo"]
```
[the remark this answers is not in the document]
spacecodeur: indeed, I didn't see turbo in the dev dependencies ! I'll fix that

spacecodeur:
> > I can't run the web server for the documentation website (the tuono dev command seems work well, but a curl http://127.0.0.1:3000 from container failed). I haven't found how to fix this problem yet..
> 
> Is the error something like this?
> 
> ```
> thread 'main' panicked at crates/tuono/src/cli.rs:111:18:
> Failed to build tuono.config.ts: Os { code: 2, kind: NotFound, message: "No such file or directory" }
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

Its not this error, I indeed run the `pnpm install` command from the `.../app/documentation` dir then the `tuono dev` command from the same dir. (I work only from the container here)
No error occured, my terminal shows : 

  ⚡ Tuono v0.16.9
  Ready at: http://localhost:3000
  
But if I open my browser from my host machine, I get the error : 

image

Maybe its about a bad config with the network and docker compose. So, in another terminal from the container, I tried to run a `curl http://127.0.0.1:3000` command directly from the docker container and I get the error : 

curl: (7) Failed to connect to 127.0.0.1 port 3000 after 0 ms: Couldn't connect to server

Currently I stoped my investigations with theses two commands from the container : 

- the `ps aux | grep tuono` : 

```
root        1068  0.0  0.0 2256996 15784 pts/1   Sl+  Jan01   0:00 tuono dev
root        1123  0.4  0.2 22635660 136100 pts/1 Sl+  Jan01   0:00 node node_modules/.bin/../tuono/bin/watch.js
root        1223  0.0  0.0 724468 15388 pts/1    Sl+  Jan01   0:00 /tuono/node_modules/.pnpm/@esbuild+linux-x64@0.21.5/node_modules/@esbuild/linux-x64/bin/esbuild --service=0.21.5 --ping
root        1535  0.0  0.0   3324  1456 pts/2    S+   00:00   0:00 grep tuono
```

- and the `netstat -anp` command : (I installed manually netstat for debug ;) )

```
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 172.17.0.2:3000         172.17.0.1:53222        TIME_WAIT   -                   
tcp6       0      0 ::1:3001                :::*                    LISTEN      1123/node           
Active UNIX domain sockets (servers and established)
Proto RefCnt Flags       Type       State         I-Node   PID/Program name     Path
unix  3      [ ]         STREAM     CONNECTED     513913   1123/node            
unix  3      [ ]         STREAM     CONNECTED     513915   1123/node            
unix  3      [ ]         STREAM     CONNECTED     513916   1223/esbuild         
unix  3      [ ]         STREAM     CONNECTED     513914   1223/esbuild         
unix  3      [ ]         STREAM     CONNECTED     520744   1068/tuono           
unix  3      [ ]         STREAM     CONNECTED     520745   1068/tuono
```

the processus `tuono dev` is running, but the port 3000 seems not to be well used.. I'll continue to investigate later :)

docker/compose.yml
```diff
@@ -0,0 +1,22 @@
...
+      context: ../
+      dockerfile: ./docker/Dockerfile
+      args:
+        RUST_VERSION: 1.83
+        BASE_OS: bookworm
+        NODE_VERSION: 22.11.0
+    image: tuono-source-image
+    ports:
+      - "3000:3000" # frontend 
+      - "3001:3001" # backend
```
Valerioageno, line 15: I think we can remove this port binding. The port `3001` should never be accessed by the client.

> Note: port `3001` is the server exposed by vite. The client interact with it by calling `localhost:3000/vite-server/*`. Our dev server then just proxies all the requests to vite.

Valerioageno:
Thanks for this! This is a great addition.
Maybe later I'll take a look at the port binding issue! 😄

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+ENV NVM_DIR /usr/local/nvm
+
+RUN ["apt", "update"]
+RUN ["apt", "install", "curl", "-y", "--no-install-recommends"]
+
+RUN ["curl", "https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh", "-o", "/tmp/install_nvm.sh"]
+RUN ["bash", "/tmp/install_nvm.sh"]
+
+ARG NODE_VERSION
+RUN ["sh", "-c", ". /usr/local/nvm/nvm.sh && nvm install ${NODE_VERSION}"]
```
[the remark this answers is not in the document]
spacecodeur: The more I think about it, the more I realize that building an image with an embedded version manager (nvm) is an anti-pattern in Docker. [This post](https://softwareengineering.stackexchange.com/a/378598) illustrates my point quite well
The best approach would be for Docker to know which version of Node it needs to install and use directly during the image build.

A good solution, in my opinion, would be to inject the wanted version [like this](https://stackoverflow.com/a/62059571/2137454) and, during the container building, download the standalone Node version that we need. 

It's straightforward to do and avoids having to install and use nvm, saving time during the container build process.

WDYT ?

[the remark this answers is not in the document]
marcalexiei: I'm not a docker expert, but from the resources you have linked it seems a more correct approach!

This makes me also wonder if something similar should be performed also for rust. 
Maybe a toolchain file could be useful for this kind of operation?

https://rust-lang.github.io/rustup/concepts/toolchains.html

docker/compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    build:
+      context: ../
+      dockerfile: ./docker/Dockerfile
+      args:
+        RUST_VERSION: 1.83
+        BASE_OS: bookworm
+        NODE_VERSION: 22.11.0
+    image: tuono-source-image
+    ports:
+      - "3000:3000" # for apps/documentation frontend app
```
marcalexiei, line 14: Actually this is incorrect:
someone could one of the site inside `examples` folder

```suggestion
      - "3000:3000" # expose default port used by tuono server
```

spacecodeur: yes indeed !

docker/Dockerfile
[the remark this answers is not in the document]
spacecodeur: Finally I adopt this approach : 
- I don't pass the NODE_VERSION var from the command line because the part `NODE_VERSION=$(cat .nvmrc | tr -cd [:digit:].)` will don't work on non-unix OS (hello Windows !). So, the `.nvmrc` file is used directly during the container building in the Dockerfile

> This makes me also wonder if something similar should be performed also for rust. Maybe a toolchain file could be useful for this kind of operation?
> 
> https://rust-lang.github.io/rustup/concepts/toolchains.html

I wasn't aware of this mechanism in Rust, it's something worth exploring in a future PR after the one here ?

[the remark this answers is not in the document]
marcalexiei: > I don't pass the NODE_VERSION var from the command line because the part NODE_VERSION=$(cat .nvmrc | tr -cd [:digit:].) will don't work on non-unix OS (hello Windows !)

Windows always brings a lot of joy as you can see in this [test file](https://github.com/tuono-labs/tuono/blob/c3fda4a3a5fc97783d9e47bfe76de4ebd2947fae/packages/tuono-lazy-fn-vite-plugin/tests/transpileSource.test.ts#L33-L41) 😅
Nevertheless I like the new approach, thank you for this change!
At least we don't have to update docker related files when updating node via `.nvmrc`!

---

> I wasn't aware of this mechanism in Rust, it's something worth exploring in a future PR after the one here ?

Fine by me, but let' see what @Valerioageno thinks about this since he is the rust expert here 😀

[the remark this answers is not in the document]
Valerioageno: Thanks for it! Definitely, fixing the rust toolchain would be a great improvement as well. Agree on doing it in a separate PR

Valerioageno:
I tested the configuration locally and everything works! ⚡ 

I noticed that I had to do again `pnpm install` on the root. Is that expected?

spacecodeur:
> I tested the configuration locally and everything works! ⚡

Yeah ! 
 
> I noticed that I had to do again `pnpm install` on the root. Is that expected?

Hmmm its not normal because this command is ran during the container building (the line `RUN ["pnpm", "install"]` at the end of the Dockerfile)

Valerioageno:
@spacecodeur Could you please check out the needed `pnpm install`? I tried again and it seems required. It is not a big deal, but it would be great to avoid that before merging!

great improvement btw!

marcalexiei:
> Once the work with Docker is complete, I can suggest updating the documentation to summarize how to use it, if you'd like :)

Also, could you please also update the documentation adding a section about docker in [Local development](https://tuono.dev/documentation/contributing/local-development)?

> [!WARNING]
> We just updated the documentation site.
> To avoid conflicts be sure to merge changes from `main` before updating that page

spacecodeur:
> > Once the work with Docker is complete, I can suggest updating the documentation to summarize how to use it, if you'd like :)
> 
> Also, could you please also update the documentation adding a section about docker in [Local development](https://tuono.dev/documentation/contributing/local-development)?
> 
> Warning
> 
> We just updated the documentation site. To avoid conflicts be sure to merge changes from `main` before updating that page

I wanted to try using "git rebase" to update my branch (I'm more familiar with "git merge"). I see that it also included in the PR commits that are not directly related to it.

I hope this won't cause any issues.

marcalexiei:
> I see that it also included in the PR commits that are not directly related to it.

That is not a rebase. It's looks more like a mass cherry pick to me.
If the rebase was made correctly there should not be 20 commits from main after the initial commits of this pr.

---

> I hope this won't cause any issues.

We are using squash as merge strategy so there shouldn't be problem since the PR is merged in a single commit

---

> I wanted to try using "git rebase" to update my branch (I'm more familiar with "git merge"

If you are more familiar with merge, then use merge, especially when working on a PR with lots of content like this one.

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    tar xf node-v${NODE_VERSION}-linux-x64.tar.xz --strip-components 1 && \
+    rm node-v${NODE_VERSION}-linux-x64.tar.xz /tmp/.nvmrc"]
+
+## Ensure Node.js binaries are available in the PATH
+
+ENV PATH /usr/local/node/bin:$PATH
+
+# pnpm installation
+
+RUN ["npm", "install", "-g", "pnpm"]
```
marcalexiei, line 30: @spacecodeur I'm unable to build the image.

From the repo root I run

```sh
docker compose -f docker/compose.yml up --build -d
```

And an error occurs:

```text
[+] Running 0/0
[+] Running 0/1o  Building                                                                                                         0.1s 
[+] Building 2.6s (10/15)                                                                                          docker:desktop-linux 
 => [tuono internal] load build definition from Dockerfile                                                                         0.0s
 => => transferring dockerfile: 1.41kB                                                                                             0.0s
 => WARN: InvalidDefaultArgInFrom: Default value for ARG rust:${RUST_VERSION}-${BASE_OS} results in empty or invalid base image n  0.0s
 => [tuono internal] load metadata for docker.io/library/rust:1.83-bookworm                                                        0.6s 
 => [tuono internal] load .dockerignore                                                                                            0.0s
 => => transferring context: 2B                                                                                                    0.0s
 => [tuono  1/11] FROM docker.io/library/rust:1.83-bookworm@sha256:a45bf1f5d9af0a23b26703b3500d70af1abff7f984a7abef5a104b42c02a29  0.0s
 => => resolve docker.io/library/rust:1.83-bookworm@sha256:a45bf1f5d9af0a23b26703b3500d70af1abff7f984a7abef5a104b42c02a292b        0.0s
 => [tuono internal] load build context                                                                                            1.8s
 => => transferring context: 11.20MB                                                                                               1.8s 
 => CACHED [tuono  2/11] RUN ["mkdir", "/usr/local/node"]                                                                          0.0s
 => CACHED [tuono  3/11] WORKDIR /usr/local/node                                                                                   0.0s
 => CACHED [tuono  4/11] COPY .nvmrc /tmp/.nvmrc                                                                                   0.0s
 => CACHED [tuono  5/11] RUN ["/bin/sh", "-c", "NODE_VERSION=$(cat /tmp/.nvmrc) &&     wget https://nodejs.org/dist/v${NODE_VERSI  0.0s
 => ERROR [tuono  6/11] RUN ["npm", "install", "-g", "pnpm"]                                                                       0.2s
------                                                                                                                                  
 > [tuono  6/11] RUN ["npm", "install", "-g", "pnpm"]:
0.105 rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
[+] Running 0/1
 ⠧ Service tuono  Building                                                                                                         2.7s 
failed to solve: process "npm install -g pnpm" did not complete successfully: exit code: 133
```

<details><summary>Log screenshot</summary>
<p>

image

</p>
</details> 

<details><summary>System info</summary>
<p>

System:
    OS: macOS 15.2
    CPU: (8) arm64 Apple M3
    Memory: 94.22 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
Virtualization:
    Docker: 27.4.0 - /usr/local/bin/docker
    Docker Compose: 2.31.0 - /usr/local/bin/docker-compose
</p>
</details> 


Could you please take a look?

spacecodeur: this warn is strange in yours logs : 

 => WARN: InvalidDefaultArgInFrom: Default value for ARG rust:${RUST_VERSION}-${BASE_OS} results in empty or invalid base image n  0.0s
 
 I take a look :)

spacecodeur: after some searches ([here](https://stackoverflow.com/questions/71040681/qemu-x86-64-could-not-open-lib64-ld-linux-x86-64-so-2-no-such-file-or-direc) by example), maybe the simple solution is to add `platform: "linux/amd64"` in the `docker/compose.yml` file like this (just below the `tuono` entry) : 

```yml
services:
  tuono:
    platform: "linux/amd64" # <= add this one here
    container_name: tuono-source-container
    hostname: tuono-source-container
    ...
```

I haven't Mac, can you tell me if this change works for you ? :) 

Make sure to delete before the container/image/... (and even Docker's cache, just to be safe—I’m always cautious, haha). I included an example command in the initial message of my PR :)

spacecodeur: the `InvalidDefaultArgInFrom: Default value for ARG rust:${RUST_VERSION}-${BASE_OS} results in empty or invalid base image n 0.0s` doesn't seem to be the problem here. 
This warn isn't important : the solution is to set a default value in the Dockerfile like this : 

```docker
ARG BASE_OS=... # where ... maybe the same value in compose.yml
ARG RUST_VERSION=... 
```

I personally prefer to keep the warning instead repeat the same data in two differents files

marcalexiei: > I haven't Mac, can you tell me if this change works for you ? :)

I was able to successfully build the docker image and successfully see and edit the documentation website

<details><summary>I have attached a screenshot to show it works perfectly!</summary>
<p>

docker-works

</p>
</details> 

----

> the InvalidDefaultArgInFrom: Default value for ARG rust:${RUST_VERSION}-${BASE_OS} results in empty or invalid base image n 0.0s doesn't seem to be the problem here.
> [...]
> I personally prefer to keep the warning instead repeat the same data in two different files

I prefer to avoid warnings in logs. 
If this means duplicating, I would duplicate the values adding a comment to keep the two files in sync.

Something similar to what I have done for pr-title-checker:

https://github.com/tuono-labs/tuono/blob/c75d83dc7c7354d7745c6fc782f0ea13f75f1ea3/.github/workflows/pr-title-checker.yml#L25-L27

https://github.com/tuono-labs/tuono/blob/c75d83dc7c7354d7745c6fc782f0ea13f75f1ea3/apps/documentation/src/routes/documentation/contributing/pull-requests.mdx?plain=1#L58-L66

I understand this is a matter of personal taste, but I would like to share my thoughts on this. 
Feel free to leave it as is, though.

marcalexiei, line 30: I just you install pnpm via `npm` this is actually installing the last version of `pnpm`  which is 9.15.2. On the project we are on 9.14.2.

This might be an issue when pnpm releases a new major version.

I think that node [corepack](https://nodejs.org/docs/latest-v22.x/api/corepack.html) could help us here:

```suggestion
RUN ["corepack ", "install"]
```

This command should install the package manager configured in the local project (`package.json#packageManager` entry)

marcalexiei (commented):
Unfortunately the "manual `pnpm install`" is still required to make the image work.
I also noticed an additional improvement that can be made to install the proper version of `pnpm`.

I have created two threads to discuss each point easily.

docker/Dockerfile
marcalexiei: Leaving this conversation open as a reminder to add `platform: "linux/amd64"` to the docker file. Since you suggested it feel free to add a comment for it! :)

wildmatthieu: > P.S.: I don't know much about docker but if args can accept a default value,
ARGs in Dockerfile can accept defaults values indeed. I think Ill do what you said : add defaults values in the Dockerfile even if they already figure in the composer.yml file and add a comment :) 

>  why not adding a default value and remove the explicit value from compose.yml?

The pro to have the compose.yml is to centralize all the global values/variables for the container building. 
If in the future we manage multiple Dockerfile (one for Tuono, another for the DB, ...) : it is more convenient to have a central file to consolidate the packages/softwares versions (sometimes shared) between the different Docker machines.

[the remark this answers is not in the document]
wildmatthieu: After some searches ( I don't reproduce yours issues :/ maybe, again, because the MAC OS system) : 
I found theses two ideas to address the issue : 

[This one](https://github.com/vitejs/vite/discussions/15532#discussioncomment-8038794) : can you confirm me this => you don't any operation beside : 
- clone the project in a new empty directory
- switch to the dockerize-repo branch
- run the build command 
? 
(by example : you don't run a command like `pnpm i` from you host machine ?)

[This another](https://github.com/vitejs/vite/discussions/15532#discussioncomment-8141236) : before the build command, can you add this optionalDependencies into the package.json ? Its just for tests and try to find at least one backup solution

spacecodeur: > P.S.: I don't know much about docker but if args can accept a default value,

ARGs in Dockerfile can accept defaults values indeed. I think Ill do what you said : add defaults values in the Dockerfile even if they already figure in the composer.yml file and add a comment :) 

>  why not adding a default value and remove the explicit value from compose.yml?

The pro to have the compose.yml is to centralize all the global values/variables for the container building. 
If in the future we manage multiple Dockerfile (one for Tuono, another for the DB, ...) : it is more convenient to have a central file to consolidate the packages/softwares versions (sometimes shared) between the different Docker machines.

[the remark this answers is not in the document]
spacecodeur: After some searches ( I don't reproduce yours issues :/ maybe, again, because the MAC OS system) : 
I found theses two ideas to address the issue : 

[This one](https://github.com/vitejs/vite/discussions/15532#discussioncomment-8038794) : can you confirm me this => you don't any operation beside : 
- clone the project in a new empty directory
- switch to the dockerize-repo branch
- run the build command 
? 
(by example : you don't run a command like `pnpm i` from you host machine ?)

[This another](https://github.com/vitejs/vite/discussions/15532#discussioncomment-8141236) : before the build command, can you add this optionalDependencies into the package.json ? Its just for tests and try to find at least one backup solution

spacecodeur: Ill try this tomorrow !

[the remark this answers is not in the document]
marcalexiei: Thank you very much for your investigation!

I was running docker commands from the same folder I'm using to work on `tuono`.
Doing a new checkout, without performing any `pnpm install` on the host machine,
produces the correct image and everything works as expected!

What do you think about adding a mention about to this in the new docker documentation section?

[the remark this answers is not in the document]
marcalexiei: @Valerioageno when you have the chance, could you please check if it works on your end?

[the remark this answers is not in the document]
Valerioageno: Unfortunately, I won't be able to check this before EOW. If you tested it and it works good for me

apps/documentation/src/routes/documentation/contributing/local-development.mdx
```diff
@@ -30,12 +30,16 @@ git clone https://github.com/<your-name-here>/tuono
 cd tuono
 ```
 
-### Rust tool chain
+For the next steps, you can either set up the environment directly on your computer or use a ready-to-use Docker image (go to [Use Docker](#use-docker) part).
```
Valerioageno, line 33:
```suggestion
For the next steps, you can either set up the environment directly on your computer or use a ready-to-use Docker image (go to the [Docker setup](#docker-setup) part).
```

apps/documentation/src/routes/documentation/contributing/local-development.mdx
```diff
@@ -30,12 +30,16 @@ git clone https://github.com/<your-name-here>/tuono
...
+
+### Set up your computer directly
```
Valerioageno, line 35:
```suggestion
### Local setup
```

apps/documentation/src/routes/documentation/contributing/local-development.mdx
```diff
@@ -65,6 +69,45 @@ pnpm run check-all
 cargo build
 ```
 
+### Use Docker
```
Valerioageno, line 72:
```suggestion
### Docker setup
```

apps/documentation/src/routes/documentation/contributing/local-development.mdx
```diff
@@ -65,6 +69,45 @@ pnpm run check-all
...
+
+#### Introduction
+
+Docker takes care of configuring the development environment, so you don’t need to worry about installing specific versions of packages like Node.js or pnpm. The Docker image handles this for you!
+Using Docker, the only operations you’ll need to manage on your host machine are related to git.
+Some IDEs (such as vscode with the [Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)) allow you to connect directly to the Docker container.
+
+#### Build the Docker Container for Tuono
```
Valerioageno, line 80:
```suggestion
#### Build the Tuono's Docker Container
```

apps/documentation/src/routes/documentation/contributing/local-development.mdx
```diff
@@ -65,6 +69,45 @@ pnpm run check-all
...
+
+First, ensure Docker is installed on your machine by following the instructions on the [Docker official site](https://docs.docker.com/engine/install/).
+
+Once Docker is installed, use the following command to build the image into a container named tuono-source-container:
+
+```sh
+# Note: Execute this command at the project root
```
Valerioageno, line 87:
```suggestion
# Note: Execute this command in the project root
```

apps/documentation/src/routes/documentation/contributing/local-development.mdx
```diff
@@ -65,6 +69,45 @@ pnpm run check-all
...
+docker compose -f docker/compose.yml up --build -d
+```
+
+To verify that everything is working as expected, run the following command:
+
+```sh
+docker images && docker ps -a --size
+```
+
+Or on Windows :
```
Valerioageno, line 97:
```suggestion
Or on Windows:
```

apps/documentation/src/routes/documentation/contributing/local-development.mdx
```diff
@@ -65,6 +69,45 @@ pnpm run check-all
...
+
+```powershell
+docker images; docker ps -a --size
+```
+
+You should see the image named tuono-source-image and the container tuono-source-container. The container's status should be "Up".
```
Valerioageno, line 103:
```suggestion
You should see the image named tuono-source-image and the container tuono-source-container. The container's status should be `up`.
```

docker/Dockerfile
spacecodeur: Features I didn't know. thank for share it :D

marcalexiei (dismissed):
LGTM!

I prefer to wait that @Valerioageno test the image on a Windows machine before merging (just to be sure).

spacecodeur:
> @spacecodeur documentation code formatting CI is failing. You should run `pnpm docs:format` and submitting the changes.

Ill take a look soon

spacecodeur:
> LGTM!
> 
> I prefer to wait that @Valerioageno test the image on a Windows machine before merging (just to be sure).

Already tested on Windows! But two tests are always better than one.

marcalexiei:
@spacecodeur is possibile to test the image building via GitHub actions?
This way we can test image is building correctly across unix-like and windows platforms and we don't introduce bugs with in the future.

spacecodeur:
@marcalexiei 
hmmm I'm not familiar with github action , but I can take a look with pleasure
I’ll probably reach out to you on Discord before the end of the week :)

Valerioageno:
Hey guys! I'm travelling right now. I expect to test it the first days of the next week!

Valerioageno:
On windows' Docker I get the following error.

Screenshot 2025-01-11 195714

Works like a charm when I use WLS and run Docker from within

Valerioageno:
Probably it is just a problem with my dirty containers setup. I use WLS and I tried to install Docker directly on windows. Maybe something went wrong with WLS. If windows works for you we can merge.

Just fix the CI pipelines!

marcalexiei (approved):

PR #279, opened 2025-01-02
chore(renovate): setup a package rule for eslint

marcalexiei:
## Context & Description

Having a single PR for each dependency might be a pain when we have to update.
Using [packageRules](https://docs.renovatebot.com/configuration-options/#packagerules)  should help mitigate this problem.

At the moment I only a group for `eslint` related packages. 
Do you have any idea for additional groups?


Valerioageno (approved):
Very cool!

Issue #242, opened 2024-12-22
[Bug]: Linking appears broken in catch-all file system routes


pveierland:
### Description

When using catch-all file system routes, using `<Link>` appears to not work correctly.

See example code here: https://github.com/pveierland/tuono/tree/issue_catch_all/examples/tuono-app/src/routes

## Test case 1: Routing without catch-all

1. Run `tuono dev`
2. Open `http://localhost:3000/pokemon_catch_one/blastoise` in browser.
✅ Page is correctly loaded.
3. Click `blastoise` link.
✅ Page is correctly rendered.
4. Click `charmander` link.
✅ Page is correctly loaded.
5. Click `charmander` link.
✅ Page is correctly rendered.
6. Click `blastoise` link.
✅ Page is correctly loaded.

## Test case 2: Routing with catch-all

1. Run `tuono dev`
2. Open `http://localhost:3000/pokemon_catch_all/blastoise` in browser.
✅ Page is correctly loaded.
3. Click `blastoise` link.
✅ Page is correctly rendered.
4. Click `charmander` link.
❌ Page is stuck with `{ data: undefined, isLoading: false` and Rust endpoint not called.
5. Open `http://localhost:3000/pokemon_catch_all/charmander` in browser.
✅ Page is correctly loaded.
6. Click `charmander` link.
✅ Page is correctly rendered.
7. Click `blastoise` link.
❌ Page is stuck with `{ data: undefined, isLoading: false` and Rust endpoint not called.

### Expected behaviour

The expected behavior is that links in catch-all file system routes would work normally.

### How to reproduce

See description.

### System Info

```shell
System:
    OS: Linux 6.6 NixOS 24.05 (Uakari) 24.05 (Uakari)
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-1370P
    Memory: 47.59 GB / 62.45 GB
    Container: Yes
    Shell: 5.2.32 - /run/current-system/sw/bin/bash
  Binaries:
    Node: 20.18.1 - /nix/store/wfxq6w9bkp5dcfr8yb6789b0w7128gnb-nodejs-20.18.1/bin/node
    Yarn: 1.22.22 - /nix/store/jcgdksj946l5l42c2y9ks2l4g6n74h3f-yarn-1.22.22/bin/yarn
    npm: 10.8.2 - /nix/store/wfxq6w9bkp5dcfr8yb6789b0w7128gnb-nodejs-20.18.1/bin/npm
  npmPackages:
[...]

PR #287, opened 2025-01-04
fix: catch-all server side handler

Valerioageno:
## Context & Description

Close: https://github.com/tuono-labs/tuono/issues/242


marcalexiei (approved):
I tested this change by porting the example from the issue locally and tested following the provided steps:
Now everything works as expected!

PR #307, opened 2025-01-13
docs(configuration): fill page content

### LINKED ISSUE #303: [Documentation]: add `tuono.config.ts` reference
The page is https://tuono.dev/documentation/configuration
The source is `apps/documentation/src/routes/documentation/configuration.mdx`

avj2352:
## Context & Description

Close #303

Fill [tuono.config.ts](https://tuono.dev/documentation/configuration) with `vite`  references.

### DIFF
--- a/apps/documentation/src/routes/documentation/configuration.mdx
+++ b/apps/documentation/src/routes/documentation/configuration.mdx
@@ -11,4 +11,142 @@ import Breadcrumbs, { Element } from '@/components/breadcrumbs'
 
 # Configuration
 
-TODO
+`tuono.config.ts` is a configuration file used in Tuono framework to customize aspects of the 
+build process and development environment. Tuono leverages `Vite` - which is a modern, 
+fast build tool and development server for web applications.
+
+The `TuonoConfig` interface serves as a wrapper for `vite.config.ts` -
+
+```typescript
+// ->> node_modules/tuono/dist/esm/config/types.d.ts
+export interface TuonoConfig {
+    vite?: {
+        alias?: AliasOptions;
+        optimizeDeps?: DepOptimizationOptions;
+        plugins?: Array<Plugin>;
+    };
+}
+```
+
+As of the current version, it supports the following properties 
+
+- `alias`: Useful for configuring custom source alias paths during imports, similar to `Next.js`. [ 👉 Read more here](https://vite.dev/config/shared-options)
+- `optimizeDeps`: Vite's dependency optimizer used only during dev. [ 👉 Read more here](https://vite.dev/config/dep-optimization-options)
+- `plugins`:  Useful for extending vite's functionality. [ 👉 Read more here](https://vite.dev/guide/using-plugins)
+
+We will cover one of the properties `alias` here in detail
+
+## 1. Configure Path Alias in Tuono
+
+After configuring aliases, you can use them in your imports (very similar to `Next.js` framework):
+
+```typescript
+// Instead of
+import MyComponent from '../../../components/MyComponent'; // relative to src folder
+
+// You can use
+import MyComponent from '@/components/MyComponent'; // considers 'src' as baseFolder
+```
+
+To achieve this, you need to add the following configuration to your `tuono.config.ts`
+
+```typescript
+// ->> tuono.config.ts
+import type { TuonoConfig } from "tuono/config";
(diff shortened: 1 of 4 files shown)


apps/documentation/src/routes/documentation/configuration.mdx
```diff
@@ -11,4 +11,142 @@ import Breadcrumbs, { Element } from '@/components/breadcrumbs'
...
+
+As of the current version, it supports the following properties 
+
+- `alias`: Useful for configuring custom source alias paths during imports, similar to `Next.js`. [ 👉 Read more here](https://vite.dev/config/shared-options)
+- `optimizeDeps`: Vite's dependency optimizer used only during dev. [ 👉 Read more here](https://vite.dev/config/dep-optimization-options)
+- `plugins`:  Useful for extending vite's functionality. [ 👉 Read more here](https://vite.dev/guide/using-plugins)
+
+We will cover one of the properties `alias` here in detail
+
+## 1. Configure Path Alias in Tuono
```
marcalexiei, line 39: This section should be rewritten to be more more linear:

1. Explain how to configure the alias in tuono.config.ts (use `@` example)
2. Add a note saying something like "Since we are using node API we should install `@types/node` to have proper type checking from `node` imports
3. Explain that the alias configuration must be aligned with `tsconfig` `compilerOptions` related properties.
4. add the snippet showing the diff between imports with relative path and imports with an alias.

apps/documentation/src/routes/documentation/configuration.mdx
```diff
@@ -11,4 +11,142 @@ import Breadcrumbs, { Element } from '@/components/breadcrumbs'
...
+```
+
+
+With these you're good to go! ✨
+
+## Practical Usecases
+
+### 01. Design libraries - ShadCN
+
+You can use libraries like `ShadCN` along with Tuono with this feature. Just follow their instruction on `Vite` site  upto `Step 03`. [👉 Link provided here](https://ui.shadcn.com/docs/installation/vite)
```
marcalexiei, line 112:
```suggestion
You can use libraries like `ShadCN` along with Tuono with this feature:
[Follow `ShadCN` instructions on `Vite` site up to `Step 03`](https://ui.shadcn.com/docs/installation/vite). 
```

apps/documentation/src/routes/documentation/configuration.mdx
```diff
@@ -11,4 +11,142 @@ import Breadcrumbs, { Element } from '@/components/breadcrumbs'
...
+
+### 02. More Path Aliases
```
marcalexiei, line 114: IMHO this section doesn't add much to what has been already written:
you already explained how a single alias can be useful so it's pretty straightforward that other aliases can be added.

Valerioageno (commented):
I think this page is too verbose.

I think the best it to reduce all the content to just a list with a quick description, default value, etc ([vite example](https://vite.dev/config/shared-options.html#resolve-alias)).

I'd put example only on very esoteric configuration entry (that IMHO we don't have yet).

marcalexiei (commented):
Let's go ahead with something like:

\# Configuration

short desc + snippet to display where to import `TuonoConfig`

\## vite

Mention about that this is a wrapper to **some** of vite options

\### vite.option1
short desc

\### vite.option2
short desc

\### vite.option3
short desc

avj2352:
I made the changes as per review comments

@marcalexiei I used the template you described above, with just one additional section - "Using Vite alias for Path configuration"

Let me know if that section is required, else I will remove that from documentation.

marcalexiei (commented):
> I used the template you described above, with just one additional section - "Using Vite alias for Path configuration"

I replaced this section with a note inside "vite.alias" section.
This note informs the user that a working example can be found in tuono-tutorial template and in the documentation code.

In the future, if users encounter trouble configuring the alias property, I would consider adding a section similar to the one you proposed on a standalone page, turning the 'Configuration' nav option into an accordion, like 'Getting Started,' 'Tutorial,' and others.

----

### Additional changes

- I noticed a problem in the MDXProvider configuration:
   (bold item were printed inside a `<div />`  tag so I replaced it with a `span`
- reword and format content
- applied prettier format

marcalexiei (approved):
Look good to me. 
Thanks for taking care of this!

PR #325, opened 2025-01-13
docs: Add useRouter Documentation

jacobhq:
## Context & Description

See [this discord thread](https://discordapp.com/channels/1307411506735874208/1328297481343336469). This PR adds documentation for the `useRouter` hook, including an example, documentation for the returned router object, and all methods returned (so far only router.push).

I like to rebase and squash PRs into one commit because I think it keeps history cleaner, so it might be a bit messy until I do that :)

### DIFF
--- a/apps/documentation/src/routes/documentation/hooks/use-router.mdx
+++ b/apps/documentation/src/routes/documentation/hooks/use-router.mdx
@@ -11,7 +11,7 @@ import Breadcrumbs, { Element } from '@/components/breadcrumbs'
 
 # useRouter
 
-The useRouter hook provides access to various data about the current route. It returns a `UseRouterResult` interface, which can be used to access route data, and to navigate between pages, as seen in the below example:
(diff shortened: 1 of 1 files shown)


Valerioageno (commented):
I left some comments. Happy to hear your thoughts on them!

P.S. No worries for squashing your commits. They are automatically squashed on PR merge

jacobhq:
Thanks for your comments, I've implemented them, and they've let me make that example much simpler!

Valerioageno:
Thanks for addressing this!

marcalexiei:
Regarding https://github.com/tuono-labs/tuono/pull/325#issuecomment-2587714884

> pathname doesn't use camel case

pathname is the standard name so it should remain the same.
Refer to the following links for more information:
- https://nodejs.org/api/url.html#url-strings-and-url-objects
- https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

----

> Return type of query (Record<string, unknown>) is actually Record<string, string> I think

You can submit a separate PR for addressing this type refine. Thanks!

Issue #258, opened 2024-12-24
[Feature request]: Add `ssr` option to `dynamic`


pveierland:
### Is your feature request related to a problem? Please describe.

The `dynamic` feature currently does not support disabling server-side rendering. This means that any component must be renderable server-side. It would be helpful if tuono had the `ssr` option in the same way as Next.js where server-side rendering can be disabled for a dynamically imported component.

This would make it easier to use components like `react-markdown` that uses browser APIs like `document` directly and for which polyfills are difficult and/or not the desired approach.

### Describe the solution you'd like

I'd like the `dynamic` function to support a second `options` object parameter with an `ssr` key which can be set to `false` to disable server-side rendering of a dynamically included component, see:

https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading#with-no-ssr

## Dependencies
- `yarn add react-markdown`

## Attempt 1: `tuono.dynamic`
```
import type { JSX } from 'react'
import { Suspense } from 'react'
import { dynamic } from 'tuono'

const Markdown: any = dynamic(() => import('react-markdown'))

export default function IndexPage(): JSX.Element {
  return (
    <Suspense>
      <Markdown>{'# Hi, *Pluto*!'}</Markdown>
    </Suspense>
  )
}
```
❌ Result: Server failure (no server log output)
## Attempt 2: `React.lazy`
```
import type { JSX } from 'react'
import { lazy, Suspense } from 'react'

const Markdown = lazy(() => import('react-markdown'))

export default function IndexPage(): JSX.Element {
  return (
    <Suspense>
      <Markdown>{'# Hi, *Pluto*!'}</Markdown>
    </Suspense>
  )
```
❌ Server log: `<unknown>:304616: Uncaught ReferenceError: document is not defined`

Valerioageno:
That is a great suggestion! Definitely we will address this soon!

PR #331, opened 2025-01-13
feat: add support for client only lazy loading with `dynamic`

Valerioageno:
## Context & Description

Closes: https://github.com/tuono-labs/tuono/issues/258

This PR extends the usage of the `dynamic` function, making its usage the same as Next.js

Quick takeaways:
1. The `tuono-lazy-fn-vite-plugin` gets completely dropped
2. `dynamic` takes an obj as second argument `{ ssr: boolean, loading: JSX.Element }`
3. The whole implementation is heavily inspired by the Next.js implementation

Good to know:
using `ssr: true` we get the following console error
Screenshot 2025-01-13 211436
This is expected because what is rendered on the server doesn't match the client.

After investigating how Next.js fixed this issue, I found that they have an error boundary that enhances and filters the console errors triggered by React. Basically, every `dynamic` hydration error gets labeled by `next/dynamic` so they automatically remove it at runtime!

I'd propose to address the `ErrorBoundary` after this and release the dynamic with the console error (which is fine since it is not blocking for the users)


packages/tuono/src/dynamic/route_lazy_loading.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react'
+import type { ReactElement } from 'react'
+
+import type { RouteComponent } from 'tuono-router'
+
+type ImportFn = () => Promise<{ default: RouteComponent }>
+
+export const __tuono__internal__lazyLoadRoute = (
```
marcalexiei, line 8: How about rename the file to `RouteLazyLoading`  and give the internal name via export: 

```ts
export { RouteLazyLoading as __tuono__internal__lazyLoadRoute } from './dynamic/route_lazy_loading'
```

packages/tuono/src/dynamic/dynamic.tsx
```diff
@@ -0,0 +1,91 @@
...
+  // This will only be rendered on the server side
+  function NoSSRLoading(): React.JSX.Element {
+    return <Loading />
+  }
+  return NoSSRLoading
+}
+
+const Loadable = <T = object,>(
+  options: LoadableOptions<T>,
+): React.ComponentType<T> => {
```
marcalexiei, line 56: The code highlight here is broken and also the syntax is look  weird.
Together they make the function hard to read:

image

I would consider using replacing the arrow function with a regular function:
```suggestion
function Loadable<T = object>(
  options: LoadableOptions<T>,
): React.ComponentType<T> {
```

packages/tuono/src/dynamic/dynamic.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * This function lets you dynamically import a component.
+ * It uses [React.lazy()](https://react.dev/reference/react/lazy) with [Suspense](https://react.dev/reference/react/Suspense) under the hood.
+ */
+export const dynamic = <T = object,>(
+  importFn: Loader<T>,
+  opts?: DynamicOptions,
+): React.ComponentType<T> => {
```
marcalexiei, line 86: Same comment of `Loadable`

packages/tuono/src/dynamic/route_lazy_loading.tsx
marcalexiei, line 8: Additional note:
If the file exports a component named `RouteLazyLoading` the file should be have the same name.
For me using underscore_case for file housing React components is strange.

On `apps/documentation/src/components` they use kebab-case., 
but maybe there is a reason about the different naming casing between component file and function I don't know.

My opinion is that the file housing React components should have the same name of the component function, so they should use camelCase. 

WDYT?

marcalexiei (commented):
Few additional considerations:

- It would be useful to add some unit tests about this component for both client and ssr case (we should be able to mock window via `vitest`).
- About
   > release the dynamic with the console error (which is fine since it is not blocking for the users)

  We should add the documentation about `dynamic()` usage alongside a warning about the expected error (informing that it will be fixed in the next release)

packages/tuono/src/dynamic/route_lazy_loading.tsx
Valerioageno: Agree. We should keep consistency and `PascalCase` is the best IMO

marcalexiei: #348

marcalexiei (approved):

PR #335, opened 2025-01-13
docs: more cohesive documentation for linking and navigation, `Link` and `useRouter` 

jacobhq:
## Context & Description

I've written the docs for the `Link` component, and I've updated the docs for linking and navigation to provide more info to those choosing between the routing methods we are providing. I still need to proof-read a little bit, so it's a draft PR.


Valerioageno:
This PR is set as "draft". Can it be reviewed?

apps/documentation/src/routes/documentation/components/link.mdx
```diff
@@ -11,4 +11,27 @@ import Breadcrumbs, { Element } from '@/components/breadcrumbs'
...
+```
+
+> Additional considerations:
+>
+> - Any `<Link />` in the viewport is prefetched by default when it appears within the viewport.
+> - The corresponding data for server-rendered routes is fetched only when the `<Link />` is clicked.
+
+## Reference
+
+The `Link` component excepts the below optional props, as well as all those accepted by the HTML `a` tag.
```
Valerioageno, line 31: I would rephrase this a little bit since `href` is not optional. WDYT?

apps/documentation/src/routes/documentation/components/link.mdx
```diff
@@ -11,4 +11,27 @@ import Breadcrumbs, { Element } from '@/components/breadcrumbs'
...
+
+- `preload`: `boolean` - Whether to preload the page if the link is within the viewport. Defaults to `true`.
+- `scroll`: `boolean` - Whether to preserve scroll position between navigations. Defaults to `true`.
+- `href`: `string` - Relative path to the page in your application. Anchor names (links with hashtags) are also valid.
+
+The `Link` component uses [`router.push`](/documentation/hooks/use-router) under the hood.
```
Valerioageno, line 37: I'm not sure this is meaningful for the users

apps/documentation/src/routes/documentation/components/link.mdx
Valerioageno, line 1: I would mention somewhere that for links that point outside the application, the user should use the `<a/>` tag. WDYT?

apps/documentation/src/routes/documentation/routing/link-and-navigation.mdx
```diff
@@ -16,7 +16,14 @@ The Tuono router facilitates client-side route transitions between pages, enhanc
 - Enabling smooth navigation in single-page applications (SPAs)
 - Preventing page reloads during site navigation
 
-How? Using the `Link` component provided from Tuono exports:
+Tuono provides two choices for navigation between pages:
+
+- The `Link` component - to be used in place of an `a` tag
+- The `useRouter` hook - to be used programmatically change routes, or to access data about the current route
```
Valerioageno, line 22: Sorry, I'm not an English native speaker. Isn't this better?

```suggestion
- The `useRouter` hook - to be used programmatically to change routes, or to access data about the current route
```

apps/documentation/src/routes/documentation/routing/link-and-navigation.mdx
```diff
@@ -44,17 +51,14 @@ The example above uses multiple links. Each one maps a path (`href`) to a known
 - `/about` → `src/routes/about.tsx`
 - `/blog/hello-world` → `src/routes/blog/[slug].tsx`
 
-> Additional considerations:
->
-> - Any `<Link />` in the viewport is prefetched by default when it appears within the viewport.
-> - The corresponding data for server-rendered routes is fetched only when the `<Link />` is clicked.
+> For more information about this component visit the [dedicated API reference page](/documentation/components/link).
```
Valerioageno, line 54: Better for website accessibility:

```suggestion
> For more information about this component visit the [Link component reference page](/documentation/components/link).
```

apps/documentation/src/routes/documentation/routing/link-and-navigation.mdx
jacobhq: You're right - I think the wording of both those bullet points is a bit jumbled. I've reworded them, let me know what you think!

apps/documentation/src/routes/documentation/components/link.mdx
Valerioageno: I'm not sure. I think they might be weird for someone not very in the Typescript thing. WDYT @marcalexiei ?

jacobhq: That's a good point. Another option is to make a table like they do in the [nextjs docs](https://nextjs.org/docs/app/api-reference/components/link). Or we could just have one list of optional, and one of required.

marcalexiei: We could move `href` prop at the beginning of the list and reformat the list to look something like this:

The `Link` component behaves similarly to the `<a />` HTML tag, with a few additional features controlled by the following props:

- `href` (`string`): Required. The relative path to the page in your application. Anchor links (with hashtags) are also supported.
- `preload` (`boolean`) Default: `true`. Whether to preload the page when the link is within the viewport.
- `scroll` (`boolean`) Default: `true`. Whether to preserve the scroll position between navigations.

---

What do you think?

Valerioageno: Like this one! 
The table is a component that is not implemented yet within our MDX templating system but we could definitely do that!

jacobhq: I've implemented it. It's a good idea, but I think for the long term a table is easier to read quickly (although out of scope of this PR). I've put in a screenshot - I don't really know, but so many inline code blocks just looks a bit weird.

image

Valerioageno (dismissed):
LGTM! Thanks

apps/documentation/src/routes/documentation/components/link.mdx
Valerioageno: Definitely agree! We should improve the readability of the documentation!
Off topic but I was thinking in the short term to implement also a Table of content on the right side of each page

jacobhq:
Sorry forgot to add all my files. Ok I'm happy with this one.

marcalexiei:
[Documentation format, lint and types CI is failing](https://github.com/tuono-labs/tuono/actions/runs/12773961775/job/35607024756?pr=335)

Could you please merge latest changes from `main` and see if the (weird) error is present?

Valerioageno (approved):

PR #342, opened 2025-01-14
docs(configuration): add types and reformat `vite` properties

marcalexiei:
## Context & Description

Added types and vite documentation for each vite option.

> [!NOTE]
> `vite.optimizeDeps` doesn't have a type because it's very complex.
> Note that this option has a standalone page inside vite documentation. 

<details><summary>Page preview</summary>
<p>

image

</p>
</details>

### DIFF
--- a/apps/documentation/src/routes/documentation/configuration.mdx
+++ b/apps/documentation/src/routes/documentation/configuration.mdx
@@ -36,14 +36,24 @@ As a result, **some** Vite configuration properties can be set in the `tuono.con
 
 ### vite.alias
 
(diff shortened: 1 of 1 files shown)


Valerioageno:
WDYT about adding also the defaults?

marcalexiei:
All defaults are `undefined`.
Usually when the default value of an option is `undefined`  it is not displayed in the doc.
Refer to the documentation of the `vite` options we are using right now.
- https://vite.dev/config/shared-options.html#resolve-alias
- https://vite.dev/config/dep-optimization-options.html
- https://vite.dev/config/shared-options.html#plugins

PR #349, opened 2025-01-15
docs(configuration): remove unclear doc

Valerioageno:
## Context & Description

@marcalexiei @avj2352 I removed this sentence since I think that for the user perspective, that does not know the codebase, it does not make much sense!


marcalexiei (approved):
You are right. Let's remove it.

If users encounter trouble configuring the alias property, I would consider adding a more detailed guide in a standalone page, turning the 'Configuration' nav option into an accordion, like 'Getting Started,' 'Tutorial,' and others.

PR #350, opened 2025-01-15
docs: use PascalCase for file name

### LINKED ISSUE #348: [Refactor]: React components - use PascalCase as filename for all files housing React components
Follow up of https://github.com/tuono-labs/tuono/pull/331#discussion_r1916051603

> On apps/documentation/src/components they use kebab-case.,
but maybe there is a reason about the different naming casing between component file and function I don't know.

We should use the same case for all files housing React components across the entire repository.

> [!NOTE]
> Most of the components already have a PascalCase name so this change should only affect file name and related imports.

marcalexiei:
## Context & Description

Close  #348


apps/documentation/src/components/Breadcrumbs/Breadcrumbs.tsx
marcalexiei, line 1: - I performed some of types and components to keep use `Breadcrumbs` as component name
- rewrite the logic to generate the `application/ld+json` script content to be more type friendly

apps/documentation/src/components/breadcrumbs/index.ts
```diff
@@ -1,5 +0,0 @@
-import Breadcrumbs, { BreadcrumbElement } from './breadcrumbs'
-
-export default Breadcrumbs
-
-export { BreadcrumbElement as Element }
```
marcalexiei, line 5: This export has been removed because it was never used

apps/documentation/src/components/Breadcrumbs/Breadcrumbs.tsx
marcalexiei, line 1: - I performed some of types and components to keep use `Breadcrumbs` as component name
- rewrite the logic to generate the `application/ld+json` script content to be more type friendly

Valerioageno (approved):
LGTM! Thanks

PR #290, opened 2025-01-05
ci: add pr title check to enforce a consistent git history

marcalexiei:
## Context & Description

Close #286

- add new workflow to check PR title correctness
- add new `/documentation/contributing/pull-requests` page with all relevant information
- `pull_request_template.md` and `CONRIBUTING.md` files has been updated accordingly

---

> [!TIP]
> If you need to perform test on the title, you can open a PR in the POC repo linked in the issue.

### Update `main` branch ruleset 

@Valerioageno before merging this PR you should update `main` branch ruleset by making "PR title checker" required before a PR is mergeable.

[Please refer to "Require status checks to pass" in the example POC repo](https://github.com/marcalexiei/github-actions-pr-title-checker-playground/rules/3152235?ref=refs%2Fheads%2Fmain):

image

> [!TIP]
> After you performed this a `Required` badge should appear in near the status check informing that it is required.

PR #365, opened 2025-01-17
chore: bump to v0.17.2

Valerioageno:
## Context & Description

This bump is for releasing the new `with-tailwind` template

PR #368, opened 2025-01-18
ci: redesign CI-related workflows to allow marking them as "Required” before merging a PR

### LINKED ISSUE #336: CI: redesign CI-related workflows `on` key to allow marking them as "Require status checks to pass"
Following the merging of #304 typescript CI was failing.

## Required Status checks
To avoid this kind of problem we should mark CI workflows as required (like `pr-title-checker.yml`).
As a result,  we won't be able to merge a PR if at least one workflow didn't succeed.

In order to achieve this we should remove the `paths` filter
https://github.com/tuono-labs/tuono/blob/64220f2e27cea5c8891f571bf2b5bfacefd56afa/.github/workflows/typescript.yml#L3-L21
in the required workflows so they will be always triggered on `push` and `pull_request` events.

### Required workflows:

- `typescript-ci.yml`
- `rust-ci.yml`
- `repo-root-ci.yml`

`documentation-ci.yml` will stay as it, so it will be triggered only on related documentation changes.

## Add "CI-OK" job inside each workflow.

Each required workflow will have a "ci-ok" job which will have the following results:
- success when all other jobs are successful 
- fail when at least one of the other jobs is failed or cancelled

Example snippet:

```yml
ci-ok:
  name: CI OK
  runs-on: ubuntu-latest
  if: always()
  needs: [test, check]
  env:
    FAILURE: ${{ contains(join(needs.*.result, ','), 'failure') }}
    CANCELLED: ${{ contains(join(needs.*.result, ','), 'cancelled') }}
  steps:
    - name: Check for failure or cancelled jobs result
      shell: bash
      run: |
        echo "Failure: $FAILURE - Cancelled: $CANCELLED"
        if [ "$FAILURE" = "false" ] && [ "$CANCELLED" = "false" ]; then
          exit 0
        else
          exit 1
        fi
```

### Why doing something like this?

The "ci-ok" job of each workflows can be matched from [`main` branch ruleset](https://github.com/tuono-labs/tuono/rules/930559?ref=refs%2Fheads%2Fmain) inside the "Require status checks to pass".

By doing this can we add / edit / delete jobs inside each required workflow
without the need to update the "Require status checks to pass" list.
As a side effect this will also reduce the amount of entries to the branch ruleset making it more easy to maintain.

---

> [!TIP]
> This will come in handy when this issue will be completed and their test will be added to the CI workflows
> - #253
> - #203

---

WDYT?

marcalexiei:
## Context & Description

Close #336.

> [!NOTE]
> @Valerioageno, as already done for #290,
> you need to add CI workflows `ci_ok` (labelled "ok") jobs 
> to the the list of "Require status checks to pass" in the `main` branch ruleset.
>
> The following list contains the list of workflow that should pass before we can merge a PR:
> - "Repo root CI"
> - "Rust CI"
> - "Typescript CI"

---

> [!IMPORTANT]
> This should be address before merging this pull request.


marcalexiei:
To the future me:

> [!NOTE]
> Adding a reusable action like `.github/actions/install-node-dependencies` right now isn't worth it.
> It requires  that every `ci_ok` also performs a checkout of the repository to have the action code available.
> Might be worth doing this in the future.

Valerioageno (approved):
It looks perfect! Thank you!
In order to avoid any possible issues, WDYT about merging this after `v0.17.2` release?

marcalexiei:
What kind of issue do you have in mind?
No problem though, I'll add a comment on #365

Valerioageno:
I don't have any actually! If you feel comfortable go ahead! It happened already the CI broke something when merged so I thought it could be safer that way

marcalexiei:
> If you feel comfortable go ahead! 

I can't go ahead: 
only you can add new required status check to the ruleset of `main` branch,
which is why this PR has been opened 🙂.

---

> It happened already the CI broke something when merged so I thought it could be safer that way

It happened after merging PR with failing CI or PR changing the lock file, which this one doesn't.
If an error happens after merging this PR we should address it, 
or, as a quick workaround, we should revert the changes of this PR.

Valerioageno:
Right! I did the thing in the settings. Let's merge this!

marcalexiei:
I'll merge it after dinner, probably we will need to merge main on some opened PR branches.
I'll take care of it later, thanks!

PR #374, opened 2025-01-19
refactor: add `TuonoEntryPoint`

### LINKED ISSUE #373: [Task]: add `TuonoEntryPoint` to wrap all tuono apps
Followup to https://github.com/tuono-labs/tuono/pull/371#discussion_r1921539135

Create a component that will wrap all tuono apps (both client and server)

Right now this component will just accept a `children` prop and will wrap it inside a `StrictMode`

This component should be created inside `packages/tuono/src/shared`.

Affected files:

- https://github.com/tuono-labs/tuono/blob/6be12b4e58611286c186aed064d107b253b344d8/packages/tuono/src/hydration/index.tsx#L14-L16
- https://github.com/tuono-labs/tuono/blob/6be12b4e58611286c186aed064d107b253b344d8/packages/tuono/src/ssr/server.tsx#L62-L64

jacobhq:
## Context & Description

Fixes #373. Wasn't sure if this should be labelled as a refactor or a chore.


Valerioageno:
Hey @jacobhq, sorry in advance for that back and forth! 
I thought it could be better to just compose the components in `TuonoEntryPoint` and remove children from it:

```tsx
export function TuonoEntryPoint({
  router,
  serverProps
}: TuonoEntryPointProps): JSX.Element {
  return (<StrictMode>
      <RouterProvider router={router} serverProps={serverProps as never} />
   </StrictMode>
  )
}
```

Do you mind updating it?

packages/tuono/src/shared/TuonoEntryPoint.tsx
```diff
@@ -0,0 +1,18 @@
+import { type JSX, StrictMode } from 'react'
+import { type createRouter, RouterProvider } from 'tuono-router'
+
+interface TuonoEntryPointProps {
```
jacobhq, line 4: I'm not too happy with the types I've used here. WDYT?

Valerioageno: For the `serverProps` you can use `ServerProps` from `tuono-router` (you might need to export it!)
`router` is a class in `tuono-router`. You should export it as well but just as type

jacobhq: Ok, I've committed some changes (CI will fail, because I think this will need to be released). WDYT?

marcalexiei: We are using workspaces.

Workspaces allow us to link the different packages in the repository so they can reference each other directly. Instead of having to install and manage dependencies for each package individually, we can use workspace links (like `workspace:*`) to automatically point to the correct version of a package within the same repository.

https://github.com/tuono-labs/tuono/blob/6be12b4e58611286c186aed064d107b253b344d8/packages/tuono/package.json#L101-L102

For more info please see: https://pnpm.io/workspaces

According to this CI must pass. 
If you have trouble resolving the error let me know and I'll take a look later.

packages/tuono/src/ssr/server.tsx
```diff
@@ -40,28 +40,26 @@ import { MessageChannelPolyfill } from './polyfills/MessageChannel'
...
 
 export function serverSideRendering(routeTree: RouteTree) {
   return async function render(payload: string | undefined): Promise<string> {
-    const serverProps = (payload ? JSON.parse(payload) : {}) as Record<
-      string,
-      unknown
-    >
+    const serverProps = (
+      payload ? JSON.parse(payload) : {}
+    ) as ServerProps | null
```
jacobhq, line 57: Can you help with the type of `serverProps`? I'm getting:
```
eslint: Unsafe assignment of an error typed value.
```

marcalexiei: A simple casting to ServerProps is suffice: `JSON.parse` returns `any` so there is no need for an additional `unknown` casting

marcalexiei (approved):

marcalexiei:
@jacobhq Thank you for you work on this task!

PR #394, opened 2025-01-21
docs: table of contents

### LINKED ISSUE #343: [Documentation]: Refactor for easier readability
As per https://github.com/tuono-labs/tuono/pull/335#discussion_r1915400073, implement:

- [x] #352
- [ ] #394

jacobhq:
## Context & Description

Fixes #343

- [x] Avoid layout shift
- [x] Shorten headings
- [x] Generate heading IDs that don't include `[Object object]` when using headings with code in them
- [ ] Make component SSR
- [x] Fix layout issues on todo pages
- [x] Fix scroll tracking (track by content not headings or similar)

## Device Screenshots

<details>
<summary>Mobile</summary>
localhost_3000_(iPhone XR)
</details>
<details>
<summary>Desktop</summary>
localhost_3000_
</details>


jacobhq:
- [ ] Avoid layout shift
- [ ] Shorten headings

Valerioageno:
Hey @jacobhq, thanks for it!
1. Can you add a screenshot?
2. We should try to deliver the full feature on the first try since it doesn't make much sense to deliver a buggy experience. Could you please also address the two remaining tasks you defined before the review?

jacobhq:
### Working example

https://github.com/user-attachments/assets/131f0695-92ab-4202-af61-175f3725dc5e


### Problems with heading IDs

https://github.com/user-attachments/assets/31a7044d-74dd-400a-b91b-86c3cba800c4

I will write a thing to generate heading IDs later this evening, then the PR will be ready.

Valerioageno:
Hey! Thanks for this.
I'm locally test it! I'll let you know next days!

Valerioageno:
I found a couple of bugs! Do you mind if I add a commit with some UI improvement, and then we address the bugs?

jacobhq:
Yeah sure!

Valerioageno:
This is the UI I thought for the table of content. I'm up for suggestions

Screenshot 2025-01-23 184536

Valerioageno:
I saw two bugs:
- The table of content is not SSR. This would not be a strict requirement, but loading the table of content on client side generates a Cumulative Layout shift error.
- Clicking on an item the page scrolls to the item below.
- (bonus) I think it makes sense have indentation according to the headings (can be an improvement done separately)

Valerioageno:
> However, do you think we should we should put the table of contents more on the side to keep the content centred?

I think it makes sense for now having it closer to the content
> Good point. Most of my experience is in Next.js, so things just kind of are magically SSRd. How would I do it in Tuono?

I can try on my side to see if we can do something for it
> I will fix that :)

Let me know when you have finished so I can check the SSR thing without risking conflicts :)
> Currently scroll tracking tracks closest heading box, so maybe I should do it by content between headings if possible?

Yeah I think this might be the solution
> Might handle that in a separate PR :)

Cool! ⚡

jacobhq:
https://github.com/user-attachments/assets/59ff283c-6f58-4eaf-b528-d6a5ff1a301c

I've updated the heading tracking to use `IntersectionObserver`, which I think works better :)

Feel free to take a look at doing SSR stuff now if you like.

jacobhq:
Not totally sure why CI took 14 min and then failed - `tuono build --static` worked on my machine!

Valerioageno:
I just finished refining the last CSS. 
In the end supporting SSR was too complex so I think for this iteration we can go with simply client side rendering.

@jacobhq I noticed that the `active` status does not work anymore. Can you take a look since you have more knowledge on them? (I had to change some margins that might be the cause but they are necessary to stabilize the table of content scroll)

jacobhq:
> @jacobhq I noticed that the `active` status does not work anymore

Do you mean the initial state?

Valerioageno:
I mean that when you scroll the page the table of content does not show any active link (all greyed out)

jacobhq:
https://github.com/user-attachments/assets/979e09e7-c4cc-46a4-8096-0fd2492bd1ef

This is what I'm getting at the moment. What I think you're describing only happens at the start. (Sorry if I'm misunderstanding!)

Valerioageno:
Right. I ran again the server now and it works. Maybe I had just some cache to update! Thanks for checking and sorry for the noise

Valerioageno:
@marcalexiei When you have time, could you please take a look with fresh eyes?

apps/documentation/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -1,115 +1,115 @@
...
       <div className={classes.inner}>
         <div>
-          <div className={classes.header}>
-            <IconList
-              style={{ width: rem(20), height: rem(20) }}
-              stroke={1.5}
-            />
-            <Text className={classes.title}>Table of contents</Text>
+          <Text className={classes.title} mb={8}>
+            On this page
```
marcalexiei, line 90: image

When scrolling the page "On this page" title has no spacing from it and the main header

apps/documentation/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -1,115 +1,115 @@
...
+          </Text>
```
marcalexiei, line 91: image

When scrolling the page "On this page" title has no spacing from it and the main header

apps/documentation/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -1,115 +1,115 @@
...
 
 import { getHeadings, type Heading } from './getHeadings'
 import classes from './TableOfContents.module.css'
 
-function getActiveElement(rects: Array<DOMRect>): number {
-  if (rects.length === 0) {
-    return -1
-  }
+export function TableOfContents(): JSX.Element | null {
+  const [active, setActive] = useState<number | null>(null)
```
marcalexiei, line 10: This should be named `activeHedingIndex`. 
`active` is usually used for `boolean` variables

apps/documentation/src/components/TableOfContents/TableOfContents.module.css
```diff
@@ -1,47 +1,37 @@
...
+  min-width: 220px;
+  max-width: 220px;
+  height: 100vh;
 
   &[data-with-tabs] {
     padding-top: 0;
     top: calc(var(--mantine-spacing-xl) + 60px);
   }
+
+  @media (max-width: 60em) {
```
marcalexiei, line 15: there are some mantine CSS variables. 
Maybe they should be used here instead of setting an hardcoded dimension?

image

apps/documentation/src/components/TableOfContents/TableOfContents.tsx
```diff
@@ -1,115 +1,115 @@
...
     return (): void => {
-      window.removeEventListener('scroll', handleScroll)
+      observer.disconnect()
+      window.removeEventListener('hashchange', handleHashChange)
     }
   }, [router.pathname])
 
-  if (filteredHeadings.length === 0) {
-    return null
+  const handleClick = (
```
marcalexiei, line 66: IMHO this should be named `handleHeadingClick`

marcalexiei (commented):
Left a few comments:
- 1 UI issue
- few code suggestions, 
   they are more nice to have so let me know what you think about them.

apps/documentation/src/components/TableOfContents/TableOfContents.tsx
Valerioageno: That is because we don't have any footer...

For now I think we can leave it as it is. WDYT?

marcalexiei: Sorry, but I don't understand why the footer is involved. 

<details><summary>Check the attached image</summary>
<p>

Screenshot 2025-01-25 at 19 33 09


</p>
</details>

Valerioageno: The Table of content has sticky position. Technically with the Foote it should go up entirely. Since the only footer for now is the edit button it goes up just few pixels. 

Actually we could include the edit button in the scrollable section so that nothing scrolls.

jacobhq: See the footer on https://mantine.dev/getting-started/ for example behaviour. I think we can leave as is, because it will be fixed when we add a footer, which is something we should do anyway :).

apps/documentation/src/components/TableOfContents/TableOfContents.module.css
jacobhq: CSS media queries [don't support variables in this way](https://stackoverflow.com/questions/40722882/css-native-variables-not-working-in-media-queries). I will hardcode 62em, and leave a comment for now.

PR #402, opened 2025-01-22
docs: remove "Components > Head"  entry from sidebar

Valerioageno:
## Context & Description

Leftover of #364


marcalexiei (approved):
Good catch!
I removed also the associated mdx file.

PR #413, opened 2025-01-24
refactor: create `TuonoContext`

### LINKED ISSUE #410: Create `TuonoContext`
Pending:
- #407

---

Followup of https://github.com/tuono-labs/tuono/pull/407#discussion_r1925969617

Create a clear distinction between tuono app and router specific information.

## `TuonoContext`

This context inside `tuono` package will contain the following data:

- `jsBundles`
- `cssBundles`
- `mode`
- `devConfig`
- `initialPayloadData` - to be passed to `RouterProvider` (see dedicated section)

The context provider will be mounted inside `TuonoEntryPoint`.

## Renaming `ServerProps` to `ServerPayload` and property `props` to `data`

To resolve the following ambiguous names
- sometime `serverProps` is `ServerProps['props']`
- sometime `serverProps` is `ServerProps`

E.g.: 

https://github.com/tuono-labs/tuono/blob/8fe8930deff13205a2303f1f836d1e7680af3755/packages/tuono-router/src/components/RouterProvider.tsx#L22-L23

- `ServerProps` → `ServerPayload`: on rust the data inside the structure is named `Payload` so we are using the same names.
- `ServerProps.props` → `ServerPayload.data` this is the data that is going to be exposed to the `RouteMatch`
- `ServerProps.router` → `ServerPayload.location` this is the data that is going to be exposed to the `RouteMatch`

> [!NOTE]
> This will allow to remove all react related terminology.

## Specialisation of `RouterContext`

`RouterProvider` will contain only router related properties
that will be forwarded to its Provider from `TuonoContext`.

We plan to also to remove each usage of `window.__TUONO_SSR_PROPS__` 
by providing the initial value as a provider prop.

> [!WARNING]
> Since both context providers will be mounted inside `TuonoEntryPoint` 
> we might need an additional component that reads the data from `TuonoContext`
> E.g.:
> - `TuonoEntryPoint`
>   - `TuonoContextProvider`
>     - `RouterContextProviderWrapper` (reads data from `TuonoContextProvider`) 
>        - `RouterContextProvider`

## Tasks

<details><summary>Outdated, check comment below</summary>
<p>

1. Create a PR with the following changes:
    - creates `TuonoContext`
    - renames `ServerProps` to `ServerPayload`
2. Decide if we are going to split the work into multiple pull requests
3. Specialise `tuono-router`'s RouterContext

</p>
</details> 

---

@Valerioageno feel free to directly update the issue if I have missed something 😅

marcalexiei:
## Context & Description

Close #410


packages/tuono-router/src/types.ts
```diff
@@ -5,26 +5,38 @@ export interface Segment {
...
+ *
+ * ```ts
+ * if (isLoading) {
+ *   // data is undefined
+ * }
+ *
+ * // data is TData
+ * ```
+ *
+ */
```
marcalexiei, line 36: This could be a fine addition to type checking when writing pages in tuono. 
If you agree I can open a new task and work on it in a separate PR.

packages/tuono/src/types.ts
```diff
@@ -1,3 +1,52 @@
...
+    }
+  | {
+      mode: 'Dev'
+      devServerConfig: {
+        port: number
+        host: string
+      }
+    }
+)
+*/
```
marcalexiei, line 48: Another type checking refinements that might better describe the runtime behaviour of payload from server. Also this one should be addressed in a different PR.

packages/tuono-router/src/types.ts
Valerioageno: I'm not sure if there is the need. A router that does not have a relative `.rs` file does not have any loading.

packages/tuono/src/shared/RouterContextProviderWrapper.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * This component is needed to get the data from {@link TuonoContext}
+ * since the provider is also located in {@link TuonoEntryPoint}
+ * hence the context cannot be accessed directly there
+ *
+ * @see https://github.com/tuono-labs/tuono/issues/410
+ */
+export function RouterContextProviderWrapper(
+  props: RouterContextProviderWrapperProps,
```
Valerioageno, line 19: Why not destructuring it directly in the argument?
```suggestion
  { router }: RouterContextProviderWrapperProps,
```

packages/tuono/src/shared/TuonoContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }, [serverPayload])
+
+  return <TuonoContext value={contextValue}>{children}</TuonoContext>
+}
+
+/**
+ * @warning THIS SHOULD NOT BE EXPOSED TO USERLAND
+ */
+export function useTuonoContextServerPayload(): TuonoContextValue['serverPayload'] {
+  return useContext(TuonoContext).serverPayload
```
Valerioageno, line 49: I like this!

packages/tuono/src/shared/TuonoScripts.tsx
```diff
@@ -1,18 +1,17 @@
...
 export function TuonoScripts(): JSX.Element {
-  const { serverSideProps } = useRouterContext()
+  const serverPayload = useTuonoContextServerPayload()
 
   return (
     <>
-      <script>{`window.__TUONO_SSR_PROPS__=${JSON.stringify(serverSideProps)}`}</script>
-      {serverSideProps?.mode === 'Dev' && <DevResources />}
-      {serverSideProps?.mode === 'Prod' && <ProdResources />}
+      <script>{`window.__TUONO_SERVER_PAYLOAD__=${JSON.stringify(serverPayload)}`}</script>
```
Valerioageno, line 12: How about turning this string into a constant and share it across the source?

```suggestion
      <script>{`window[payloadWindowLocation]=${JSON.stringify(serverPayload)}`}</script>
```

packages/tuono/src/types.ts
```diff
@@ -1,3 +1,52 @@
+/**
+ * Provided from the the rust server and use in ssr env
```
Valerioageno, line 2:
```suggestion
 * Provided by the rust server and use in the ssr env
```

packages/tuono/src/types.ts
Valerioageno: I like the union Idea

Valerioageno (dismissed):
Left a couple of comments but overall I think it is very good! Thanks

packages/tuono/src/types.ts
marcalexiei: https://github.com/orgs/tuono-labs/projects/2?pane=issue&itemId=95174765

packages/tuono/src/shared/TuonoScripts.tsx
marcalexiei: `SERVER_PAYLOAD_VARIABLE_NAME` added

Valerioageno (approved):
LGTM!

PR #415, opened 2025-01-25
fix(packages/tuono): build error when config contains vite plugins

### LINKED ISSUE #414: [Bug]: Failed to read tuono.config.ts
### Description

When trying to tun `tuono build` on `examples/with-tailwind` I get:
```
➜ tuono build
[CLI] Failed to read tuono.config.ts
```

### Expected behaviour

Successful build.

### How to reproduce

Run `tuono build` in `examples/with-tailwind`.

### Screenshots

_No response_

### System Info

```shell
System:
    OS: Linux 6.11 Ubuntu 24.10 24.10 (Oracular Oriole)
    CPU: (4) x64 Intel(R) Core(TM) i3-8130U CPU @ 2.20GHz
    Memory: 2.21 GB / 7.14 GB
    Container: Yes
    Shell: 5.2.32 - /bin/bash
  Binaries:
    Node: 22.13.0 - ~/.nvm/versions/node/v22.13.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.13.0/bin/npm
    pnpm: 9.15.2 - ~/.nvm/versions/node/v22.13.0/bin/pnpm
    bun: 1.1.43 - ~/.bun/bin/bun
  Browsers:
    Chrome: 131.0.6778.264
  npmPackages:
    tuono: ^0.17.4 => 0.17.4
```

### System info (Rust)

```shell
rustc 1.84.0 (9fc6b4312 2025-01-07)
cargo 1.84.0 (66221abde 2024-11-19)
tuono 0.17.4
```

### Additional context

_No response_

marcalexiei:
## Context & Description

Close #414.

The error was caused by the serialisation via structuredClone of a configuration containing a vite plugin.
More info can be found on `packages/tuono/src/build/config/create-json-config.ts`

Relevant unit tests were added to avoid regression.

### DIFF
--- a/packages/tuono/src/build/config/create-json-config.spec.ts
+++ b/packages/tuono/src/build/config/create-json-config.spec.ts
@@ -1,0 +1,32 @@
+import fs from 'fs/promises'
+
+import { describe, expect, it, vitest } from 'vitest'
+import react from '@vitejs/plugin-react-swc'
+
+import { createJsonConfig } from './create-json-config'
+
+const writeFileSpy = vitest.spyOn(fs, 'writeFile').mockResolvedValue(void 0)
+
+describe('createJsonConfig', () => {
+  const sampleConfig = { server: { host: 'h', port: 1 } }
+
+  it('should process config with only server property', async () => {
+    await createJsonConfig(sampleConfig)
+
+    expect(writeFileSpy).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.stringContaining(JSON.stringify(sampleConfig)),
+      expect.any(String),
+    )
+  })
+
+  it('should process config with plugins', async () => {
+    await createJsonConfig({ ...sampleConfig, vite: { plugins: [react()] } })
+
+    expect(writeFileSpy).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.stringContaining(JSON.stringify(sampleConfig)),
+      expect.any(String),
+    )
+  })
+})
--- a/packages/tuono/src/build/config/create-json-config.ts
+++ b/packages/tuono/src/build/config/create-json-config.ts
@@ -21,9 +21,25 @@ const CONFIG_PATH = path.join(
 function removeViteProperties(
   config: InternalTuonoConfig,
 ): Omit<InternalTuonoConfig, 'vite'> {
-  const newConfig = structuredClone(config)
-  delete newConfig['vite']
-  return newConfig
+  /**
+   * Using {@link structuredClone} cause the following errors based on runtime env:
+   * ```text
+   * node
+   * DOMException [DataCloneError]: configureServer(s){a.push(s)} could not be cloned.
+   *
+   * vitest
+   * DataCloneError: (id) => id === runtimePublicPath ? id : void 0 could not be cloned.
+   * ```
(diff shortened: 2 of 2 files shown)


packages/tuono/src/build/config/create-json-config.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    expect(writeFileSpy).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.stringContaining(JSON.stringify(sampleConfig)),
+      expect.any(String),
+    )
+  })
+
+  it('should process config with plugins', async () => {
+    await createJsonConfig({ ...sampleConfig, vite: { plugins: [react()] } })
```
Valerioageno, line 24: WDYT about removing the `react` plugin so that we are keeping it more framework-agnostic?

```suggestion
    await createJsonConfig({ ...sampleConfig, vite: { plugins: [{id: 'plugin'} as PluginOption] } })
```

Valerioageno (approved):
What can I say...
You just nailed it! ⚡

packages/tuono/src/build/config/create-json-config.spec.ts
marcalexiei: Using a simple JS object, like you proposed, won't cause any serialisation error.
Right now `react` / `tailwind` plugins return something that cannot be serialised,
which was the cause of the error which this test is aimed to prevent in the future.

Since we are not removing `react` soon I would consider investigating an alternative later.

Valerioageno: As far as I understood in the test we want to check just that vite is stripped out. There is no need of keeping the correct values

marcalexiei: No, this test is aimed to prevent regression of #414.
The issue was caused by non serialisable data inside the `vite.plugins`, 
so to avoid regression we must create the same scenario where the `vite.plugins` contains not serialisable data.

Valerioageno: What I mean is that you are not checking that the data is not serializable (and then you strip it out). You are just removing `vite` from the cloned value.
For this reason in the test I think you don't need to put non-serializable data within the `vite` object since it will always be stripped

marcalexiei: Right now there are no known data that be set to not serialisable values except `vite.plugins`.

Again, the point here is to prevent regression of #414...
The test recreates the exact scenario that was causing that issue.

Do the following steps:
1. Restore the previous code inside `removeViteProperties`
    ```ts
     const newConfig = structuredClone(config)
     delete newConfig['vite']
     return newConfig
     ```
3. Run`pnpm --filter=tuono run test` the test with `react` plugin. 
    You will experience a `DataCloneError` error (#414)

4. Replace the `react`  plugin with a plain object.
    No error is present. 

So, with your test code there is no error with the old implementation...
so I don't see any valid reason to write that code since it was working with the old implementation which was causing  #414 error

Valerioageno: I got it! Thanks

PR #420, opened 2025-01-26
fix: check if project contains `tuono.config.ts` before running dev/build

### LINKED ISSUE #393: Better error handling when `tuono dev` | `tuono build` is run in a non-tuono directory
Say I run `tuono dev` in an empty directory:
```
~\code\tmp
❯ tuono dev
The system cannot find the path specified.
thread 'tokio-runtime-worker' panicked at C:\Users\jacob\.cargo\registry\src\index.crates.io-6f17d22bba15001f\watchexec-supervisor-3.0.0\src\job\task.rs:54:17:
all branches are disabled and there is no else branch
The system cannot find the path specified.
error: could not find `Cargo.toml` in `C:\Users\jacob\code\tmp` or any parent directory
thread 'tokio-runtime-worker' panicked at C:\Users\jacob\.cargo\registry\src\index.crates.io-6f17d22bba15001f\watchexec-supervisor-3.0.0\src\job\task.rs:54:17:
all branches are disabled and there is no else branch
```
We should be able to detect this and provide a more descriptive error.

jacobhq:
## Context & Description

Fixes #393

### DIFF
--- a/crates/tuono/src/cli.rs
+++ b/crates/tuono/src/cli.rs
@@ -1,5 +1,4 @@
 use fs_extra::dir::{copy, CopyOptions};
-use std::io::Error;
 use std::path::PathBuf;
 use std::thread::sleep;
 use std::time::Duration;
@@ -47,7 +46,15 @@ struct Args {
     action: Actions,
 }
 
+fn check_for_tuono_config() {
+    if !PathBuf::from("tuono.config.ts").exists() {
+        eprintln!("Cannot find tuono.config.ts - is this a tuono project?");
+        std::process::exit(1);
+    }
+}
+
 fn init_tuono_folder(mode: Mode) -> std::io::Result<App> {
+    check_for_tuono_config();
     check_tuono_folder()?;
     let app = bundle_axum_source(mode)?;
     create_client_entry_files()?;
@@ -55,22 +62,11 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<App> {
     Ok(app)
 }
 
-fn check_for_tuono_config() -> Result<(), Error> {
-    if !PathBuf::from("tuono.config.ts").exists() {
-        return Err(Error::new(
-            std::io::ErrorKind::NotFound,
-            "Cannot find tuono.config.ts - is this a tuono project?",
-        ));
-    }
-    Ok(())
-}
-
 pub fn app() -> std::io::Result<()> {
     let args = Args::parse();
 
     match args.action {
         Actions::Dev => {
-            check_for_tuono_config()?;
             let mut app = init_tuono_folder(Mode::Dev)?;
 
             app.build_tuono_config()
@@ -81,7 +77,6 @@ pub fn app() -> std::io::Result<()> {
             watch::watch().unwrap();
         }
         Actions::Build { ssg, no_js_emit } => {
-            check_for_tuono_config()?;
             let mut app = init_tuono_folder(Mode::Prod)?;
 
             if no_js_emit {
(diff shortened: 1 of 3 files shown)


crates/tuono/src/cli.rs
```diff
@@ -54,11 +55,22 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<App> {
...
+    Ok(())
+}
+
 pub fn app() -> std::io::Result<()> {
     let args = Args::parse();
 
     match args.action {
         Actions::Dev => {
+            check_for_tuono_config()?;
             let mut app = init_tuono_folder(Mode::Dev)?;
```
Valerioageno, line 74: I'd put the check in `init_tuono_config` to avoid duplications

crates/tuono/tests/utils.rs
```diff
@@ -11,24 +11,34 @@ pub struct TempTuonoProject {
...
         env::set_current_dir(temp_dir.path()).expect("Failed to change current dir into temp_dir");
 
-        TempTuonoProject {
+        let project = TempTuonoProject {
             original_dir,
             temp_dir,
-        }
+        };
+
+        project.add_file("./tuono.config.ts");
```
Valerioageno, line 32: You should add a test case for the scenario you are addressing with this PR

crates/tuono/tests/utils.rs
```diff
@@ -11,24 +11,34 @@ pub struct TempTuonoProject {
     temp_dir: TempDir,
 }
 
+impl Default for TempTuonoProject {
```
Valerioageno, line 14: Why this? It looks like you are not using it

crates/tuono/src/cli.rs
```diff
@@ -54,11 +55,22 @@ fn init_tuono_folder(mode: Mode) -> std::io::Result<App> {
     Ok(app)
 }
 
+fn check_for_tuono_config() -> Result<(), Error> {
+    if !PathBuf::from("tuono.config.ts").exists() {
```
Valerioageno, line 59: I think the check here should be simply:

```rs
if !PathBuf::from("tuono.config.ts").exists() {
     eprintln!("Cannot find tuono.config.ts - is this a tuono project?");
     std::process::exit(1);
}
```

Valerioageno (changes requested):

crates/tuono/tests/utils.rs
jacobhq: It was suggested by clippy:

you should consider adding a `Default` implementation for `TempTuonoProject` Help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default Note: `#[warn(clippy::new_without_default)]` on by default

I wasn't entirely sure whether we needed it so I just added it. Do you think we need it?

jacobhq: Actually I don't think we need it so I removed it because in the doc it said 'users', however this is just an internal util for tests.

crates/tuono/tests/utils.rs
```diff
@@ -18,6 +18,22 @@ impl TempTuonoProject {
...
+            original_dir,
+            temp_dir,
+        };
+
+        project.add_file("./tuono.config.ts");
+
+        project
+    }
+
+    pub fn new_with_no_config() -> Self {
```
Valerioageno, line 31: I think duplicating the function is not a good pattern. I'd consider reusing it within the `new` method like:

```rs
pub fn new() -> Self {
        let project = TempTuonoProject::new_with_no_config();

        project.add_file("./tuono.config.ts");

        project
    }
```

Valerioageno (approved):
LGTM! Thanks

Issue #465, opened 2025-01-29
[Documentation]: Add indentation to Table of Contents


jacobhq:
- [ ] #475
- [x] #467

Valerioageno:
Cool thanks!

Valerioageno:
I noticed that headings that are code blocks are not indexed by the TOC. (http://localhost:3000/documentation/hooks/use-router)
Could you take also a look at them?

jacobhq:
Sure. I'll keep this as one issue, but split into two separate PRs, because I think indentation will take longer to add than getting code headings.

Issue #428, opened 2025-01-26
[Documentation]: UI issues


marcalexiei:
Issues discovered while working on #427

## Heading have no margin

Page headings have no longer any margin.
While this isn't noticeable on most documentation pages, 
it becomes evident on pages like ["Contributing > Local development"](https://tuono.dev/documentation/contributing/local-development), 
where there are many headings close together.

<details><summary>Check the attached screenshot</summary>
<p>

Old version → New version

<img width="2053" alt="Image" src="https://github.com/user-attachments/assets/11c25f1a-8393-488a-b34c-c5ae53969dfa" />

</p>
</details> 

## Link `font-size`

Links should not have the same or a smaller font size than the paragraph they are contained:

Few pages examples:
- https://tuono.dev/documentation/contributing/pull-requests:
   <img width="553" alt="Image" src="https://github.com/user-attachments/assets/b65880d7-fb65-4480-9e84-8dcf73becdf6" />
- https://tuono.dev/documentation/contributing/local-development
   <img width="307" alt="Image" src="https://github.com/user-attachments/assets/7ff0e794-c0c6-4183-aa3b-8a09375bfc63" /> 

</p>
</details>

Valerioageno:
Thanks for this! I wanted to revamp the spacing a little bit! I'm gonna work on this!

jacobhq:
@Valerioageno I've done some work on link font size in #426. You can take a look through it if you want, as link improvements may be out of scope for that PR.

PR #466, opened 2025-01-29
docs: refine  UI

Valerioageno:
### Checklist

### Related issue

Fixes [#428](https://github.com/tuono-labs/tuono/issues/428)

### Overview

- [x] heading margins
- [x] Add headings to all the sections
- [x] Add link to each heading
- [x] Reduce padding on mobile

Screenshot 2025-01-29 193235

Note: heading with `code only` text aren't recognized by the ToC. that will be fixed here https://github.com/tuono-labs/tuono/issues/465

### DIFF
--- a/apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
+++ b/apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
@@ -5,11 +5,15 @@ import type { TitleProps } from '@mantine/core'
 import { IconLink } from '@tabler/icons-react'
 import { useHover } from '@mantine/hooks'
 
-export default function MdxTitle(props: TitleProps): JSX.Element {
-  const headingId = getIdFrom(props.children)
+export default function MdxTitle({
+  children,
+  order,
+  ...rest
+}: TitleProps): JSX.Element {
+  const headingId = getIdFrom(children)
   const { hovered, ref } = useHover<HTMLHeadingElement>()
 
-  const onLinkClick: MouseEventHandler<HTMLAnchorElement> = useCallback(
+  const handleAnchorClick: MouseEventHandler<HTMLAnchorElement> = useCallback(
     (e) => {
       e.preventDefault()
       if (ref.current) {
@@ -30,28 +34,28 @@ export default function MdxTitle(props: TitleProps): JSX.Element {
   return (
     <Box
       ref={ref}
-      data-order={props.order}
+      data-order={order}
       style={{
         scrollMargin: 80,
-        marginTop: props.order === 1 ? 0 : 20,
+        marginTop: order === 1 ? 0 : 20,
       }}
-      {...props}
+      {...rest}
     >
       <Title
         data-heading={headingId}
         id={headingId}
-        data-order={props.order}
+        data-order={order}
         display="inline"
         style={{
           scrollMargin: 80,
-          marginTop: props.order === 1 ? 0 : 20,
+          marginTop: order === 1 ? 0 : 20,
         }}
-        {...props}
+        {...rest}
       >
-        {props.children}
-        {hovered && props.order !== 1 && (
+        {children}
+        {hovered && order !== 1 && (
           <Anchor
-            onClick={onLinkClick}
+            onClick={handleAnchorClick}
             display="inline-flex"
             ml={8}
             mb={2}


apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
```diff
@@ -1,15 +1,52 @@
...
+} from 'react'
+import { Title, Anchor, type TitleProps } from '@mantine/core'
+import { IconLink } from '@tabler/icons-react'
+import { useHover } from '@mantine/hooks'
 
 export default function MdxTitle(props: TitleProps): JSX.Element {
+  const { hovered, ref } = useHover<HTMLHeadingElement>()
+
+  const onLinkClick = useCallback(
+    (e: MouseEvent<HTMLAnchorElement>): void => {
```
marcalexiei, line 16:
```suggestion
  const handleAnchorClick: MouseEventHandler<HTMLAnchorElement> = useCallback(
    (e): void => {
```

apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
```diff
@@ -1,15 +1,52 @@
...
+      e.preventDefault()
+      if (ref.current) {
+        ref.current.scrollIntoView({
+          behavior: 'instant',
+          block: 'start',
+        })
+      }
+    },
+    [ref],
```
marcalexiei, line 25: This handler should also update the location hash. 
At the moment it just perform a scroll into view of the title which isn't the behaviour I would expect from clicking an anchor link near a section title.

apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
```diff
@@ -1,15 +1,52 @@
...
+        alignItems: 'center',
+        gap: 8,
+      }}
       {...props}
       id={idGen(props.children)}
-    />
+    >
+      {props.children}
+      {hovered && props.order !== 1 && (
+        <Anchor onClick={onLinkClick} h={20}>
```
marcalexiei, line 45:
```suggestion
        <Anchor onClick={handleAnchorClick} h={20}>
```

apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
```diff
@@ -1,15 +1,52 @@
...
+  type ElementType,
+  type JSX,
+  type MouseEvent,
+  type ReactNode,
+} from 'react'
+import { Title, Anchor, type TitleProps } from '@mantine/core'
+import { IconLink } from '@tabler/icons-react'
+import { useHover } from '@mantine/hooks'
 
 export default function MdxTitle(props: TitleProps): JSX.Element {
```
marcalexiei, line 12: Why not use destructing here?

```suggestion
export default function MdxTitle({ order, children, ...rest }: TitleProps): JSX.Element {
```

apps/documentation/src/routes/documentation/hooks/use-router.mdx
```diff
@@ -35,7 +37,7 @@ export default function IndexPage(): JSX.Element {
 }
 ```
 
-## `useRouter` result
```
marcalexiei, line 38: This section describes the returned  value of the `useRouter`, IMHO result is more semantic.

Using object make useRouter look like an object rather than a hook

apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
marcalexiei, line 12: Also I think that `order` should be renamed to `level` to match HTML:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements
   
   > The `<h1>` to `<h6>` [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) elements represent six levels of section headings. `<h1>` is the highest section level and `<h6>` is the lowest.

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-level#with_heading_role

apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
Valerioageno: I forgot to put it... good catch!

apps/documentation/src/components/MdxProvider/MdxTitle/MdxTitle.tsx
Valerioageno: Order is a mantine `TitleProps` field. I wouldn't change it

Issue #450, opened 2025-01-28
[Feature request]: improve `TuonoProps` type


marcalexiei:
### Is your feature request related to a problem? Please describe.

`TuonoProps` is a very ambiguous name and should describe better the runtime behaviour of route props.

### Describe the solution you'd like

- Rename `TuonoProps` into `TuonoRouteProps`. 
   Using this name will clearly identify that this type involves a Route components
- https://github.com/tuono-labs/tuono/pull/646#discussion_r1996004374
  https://github.com/tuono-labs/tuono/blob/bfd0041c6a1a14a427543f32ca80c632783004b4/packages/tuono/src/types.ts#L50-L53

  The interface should become a union type to provide to typescript the correct anatomy of the runtime value,
   which allows a   better static code analysis:

  ```ts
  export type TuonoRouteProps<T> = { 
     data: null
     isLoading: true
   } | { 
     data: T
     isLoading: false
   } 
  ``` 

### Describe alternatives you've considered

N/A

### Additional context

Related to #449

PR #473, opened 2025-01-31
docs: add SSR page

Valerioageno:
### Checklist

### Related issue

Fixes https://github.com/tuono-labs/tuono/issues/454

### DIFF
--- a/apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
+++ b/apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
@@ -13,22 +13,29 @@ import Breadcrumbs from '@/components/Breadcrumbs'
 
 ## Overview
 
-If a page uses **Server-side Rendering**, the page HTML is generated on **each request**.
+Server side rendering (SSR) allows each page to render the HTML on the server
+side by preloading the data needed for the render itself. SSR runs on each HTTP request.
 
-> By default, every page is automatically server side rendered.
+SSR is usually needed for:
+
+- SEO
+- Faster client side rendering
+- Reduced JS bundle required for the initial load
+
+## SSR on a route
 
 To enable SSR on a page you just need to create the same file with the rust extension:
 
 - `src/routes/about.tsx`
 - `src/routes/about.rs`
 
-and then create a #[tuono_lib::handler] in the rust file, 
+and then create a `#[tuono_lib::handler]` in the rust file,
 which will be the function called on each request before rendering the HTML.
 
 ## Example
 
-For example, suppose your page needs to pre-render frequently updated data
-(read from a local file). 
+Suppose your page needs to pre-render frequently updated data
+(read from a local file).
 You can write a handler which reads a JSON file and passes it to the route like below:
 
 ```rs
--- a/apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
+++ b/apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
@@ -55,4 +55,4 @@ export default function AbouteRoute({ data }: TuonoProps) {
 }
 ```
 
-> `TuonoProps<T>` takes a generic argument for statically typing `data`.
+> `TuonoProps<T>` takes a generic argument to statically type `data`.
--- a/apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
+++ b/apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
@@ -15,7 +15,7 @@ import Breadcrumbs from '@/components/Breadcrumbs'
 
 If a page uses **Server-side Rendering**, the page HTML is generated on **each request**.
 
-> Every page is automatically server side rendered by default.
+> By default, every page is automatically server side rendered.
 
 To enable SSR on a page you just need to create the same file with the rust extension:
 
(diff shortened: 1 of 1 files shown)


apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
```diff
@@ -9,6 +9,50 @@ import Breadcrumbs from '@/components/Breadcrumbs'
...
 
-# SSR
+# Server side rendering
 
-TODO
+## Overview
+
+If a page uses **Server-side Rendering**, the page HTML is generated on **each request**.
+
+> Every page is automatically server side rendered by default.
```
marcalexiei, line 18:
```suggestion
> By default, every page is automatically server side rendered.
```

apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
```diff
@@ -9,6 +9,50 @@ import Breadcrumbs from '@/components/Breadcrumbs'
...
+
+To enable SSR on a page you just need to create the same file with the rust extension:
+
+- `src/routes/about.tsx`
+- `src/routes/about.rs`
+
+and then create an `#[tuono_lib::handler]` in the rust file that will be the
+function actually called on each request before rendering the HTML.
```
marcalexiei, line 26:
```suggestion
and then create a #[tuono_lib::handler] in the rust file, 
which will be the function called on each request before rendering the HTML.
```

apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
```diff
@@ -9,6 +9,50 @@ import Breadcrumbs from '@/components/Breadcrumbs'
...
+
+## Example
+
+For example, suppose that your page needs to pre-render frequently updated data
+(read from a local file). You can write an hanlder which reads a json file and passes
+it to the route like below:
```
marcalexiei, line 32:
```suggestion
For example, suppose your page needs to pre-render frequently updated data
(read from a local file). 
You can write a handler which reads a JSON file and passes it to the route like below:
```

apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
```diff
@@ -9,6 +9,50 @@ import Breadcrumbs from '@/components/Breadcrumbs'
...
+```tsx
+// src/routes/about.tsx
+import type { TuonoProps } from 'tuono'
+
+export default function AbouteRoute({ data }: TuonoProps) {
+  // ...
+}
+```
+
+> `TuonoProps<T>` takes a generic argument for statically typing `data`.
```
marcalexiei, line 58:
```suggestion
> `TuonoProps<T>` takes a generic argument to statically type `data`.
```

apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
```diff
@@ -9,6 +9,50 @@ import Breadcrumbs from '@/components/Breadcrumbs'
...
+
+  Response::Props(Props::new(data))
+}
+```
+
+Then you will be able to read the loaded data on the route in the route's `data` props:
+
+```tsx
+// src/routes/about.tsx
+import type { TuonoProps } from 'tuono'
```
marcalexiei, line 51: I already bring this up in #450, but IMHO we should give a more semantic name to this type.

(We can address this in a different PR)

apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
```diff
@@ -9,6 +9,50 @@ import Breadcrumbs from '@/components/Breadcrumbs'
...
+```rs
+// src/routes/about.rs
+use tuono_lib::{Request, Response, Props};
+use std::fs;
+
+#[tuono_lib::handler]
+async fn my_custom_ssr_page(req: Request) -> Response {
+  let data = fs::read_to_string("my_data.json").expect("Failed to read json file");
+
+  Response::Props(Props::new(data))
```
marcalexiei, line 43: I haven't used this feature so far but I have some doubts about `Props` name usage:
I don't know how much familiar this term can be for user when dealing with http responses and with framework different from React.
Since you plan to make `tuono` framework agnostic I would consider another name.
I would use something tied to HTTP response statuses like
`OK` (200) since another option for Response is `Redirect`.
If this isn't ok here are a few additional alternatives:
- `Content`
- `Output`
- `RouteData`

(We can address this in a different PR)

apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
```diff
@@ -9,6 +9,50 @@ import Breadcrumbs from '@/components/Breadcrumbs'
...
+# Server side rendering
 
-TODO
+## Overview
+
+If a page uses **Server-side Rendering**, the page HTML is generated on **each request**.
+
+> Every page is automatically server side rendered by default.
+
+To enable SSR on a page you just need to create the same file with the rust extension:
```
jacobhq, line 20: Just a quick one - on line 18, you say by default, but here you say it needs the Rust file. Which one? Great doc though :)

Valerioageno: You right! Better being more specific! Thanks for checking

apps/documentation/src/routes/documentation/rendering/server-side-rendering.mdx
Valerioageno: Agree. I'd propose to do an investigation about the overall DX and syntax! Agree that it is pretty rough right now

Valerioageno:
@marcalexiei @jacobhq I updated the doc! let me know if it can be merged!

jacobhq (approved):
Epic!

marcalexiei (approved):

PR #478, opened 2025-02-02
test(packages/tuono-router): refine unit tests

marcalexiei:
### Checklist

### Related issue

Fixes None

### Overview

- `packages/tuono-router/src/components/RouteMatch.spec.tsx`
  - use snapshot to test produced router structure
  - as side effect of using snapshot `@testing-library/jest-dom` can be removed
- `packages/tuono-router/src/hooks/useRoute.spec.ts`
  - move module mocking at the beginning of the file
  - create `useRouterContextMock` to allow each test to customise the hook return.
    Might be useful in the future if we need to add more tests.

### DIFF
--- a/packages/tuono-router/package.json
+++ b/packages/tuono-router/package.json
@@ -51,7 +51,6 @@
   },
   "devDependencies": {
     "@tanstack/config": "0.7.13",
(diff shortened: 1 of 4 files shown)


Valerioageno (approved):
LGTM Thanks

PR #508, opened 2025-02-05
chore: refine issue templates

marcalexiei:
### Checklist

### Related issue

Fixes nothing 

### Overview

- Added new task template
- Added Discord link to issue config `contact_links`
- Remove blank issue template

> [!NOTE]
> I haven't added any documentation related template
> since we are [planning to move the documentation in a standalone repository](https://github.com/orgs/tuono-labs/projects/2/views/2?filterQuery=-status%3ADone+repo&pane=issue&itemId=95784175)


.github/ISSUE_TEMPLATE/task.yml
Valerioageno, line 1: I don't think we need this. External are not supposed to create a `task` and we should always manage them from the project

marcalexiei: Disabling blank issues is for all users, no one can bypass it, so we need this template for us.
Also consider that someone is not a member could propose something that is not a bug or a feature

Valerioageno: I don't get it. 
1. Shouldn't we always create the issue from the project? In that case we don't have to pass through it
2. I don't think users should never create something different from `bug` or `feature`. Whatever they need to create that is not one of the two I think it is not a `task` as well

Valerioageno (approved):

PR #534, opened 2025-02-08
fix: update link component to allow modifier+click to properly work, and write tests

### LINKED ISSUE #531: [Bug]: Link leading to other documentation pages aren't opened in a new tab using cmd + click
### Description

Refer to the title

### Expected behaviour

I can open links in a new tab when using cmd + click

### How to reproduce

1. Go to https://tuono.dev/documentation/routing/link-and-navigation
2. (Cmd / ctrl) + Click on a link leading to a page of the documentation site
   - "useRouter API reference page"
   - "Link component API reference page".

### Screenshots

_No response_

### System Info

```shell
Not applicable
```

### System info (Rust)

```shell
Not applicable
```

### Additional context

_No response_

jacobhq:
### Checklist

### Related issue

Fixes #531

### Overview

This PR makes the link fallback to default `a` behaviour when a modifier key is being pressed.

### DIFF
--- a/packages/tuono-router/src/components/Link.spec.tsx
+++ b/packages/tuono-router/src/components/Link.spec.tsx
@@ -1,0 +1,129 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { render, fireEvent } from '@testing-library/react'
+
+import Link from './Link'
+
+const pushMock = vi.fn()
+const preloadMock = vi.fn()
+
+vi.mock('../hooks/useRouter', () => ({
+  useRouter: (): { push: typeof pushMock } => ({ push: pushMock }),
+}))
+
+vi.mock('../hooks/useRoute', () => ({
+  useRoute: (): { component: { preload: typeof preloadMock } } => ({
+    component: { preload: preloadMock },
+  }),
+}))
+
+let intersectionObserverCallback: ((inView: boolean) => void) | undefined
+
+vi.mock('react-intersection-observer', () => ({
+  useInView: (options: {
+    onChange: (inView: boolean) => void
+  }): {
+    ref: () => void
+  } => {
+    intersectionObserverCallback = options.onChange
+    return { ref: vi.fn() }
+  },
+}))
+
+describe('Link Component', (): void => {
+  beforeEach((): void => {
+    pushMock.mockReset()
+    preloadMock.mockReset()
+    intersectionObserverCallback = undefined
+  })
+
+  it('renders with correct href and text', (): void => {
+    const { getByRole } = render(<Link href="/test">Test Link</Link>)
+    const link = getByRole('link', { name: 'Test Link' })
+
+    expect(link.getAttribute('href')).toBe('/test')
+    expect(link.textContent).toBe('Test Link')
+  })
+
+  it('calls router.push on normal click', (): void => {
+    const { getByRole } = render(<Link href="/test">Test Link</Link>)
+    const link = getByRole('link')
+
+    fireEvent.click(link)
+    expect(pushMock).toHaveBeenCalledWith('/test', { scroll: true })
+  })
+
+  it('does not navigate if href starts with "#"', (): void => {
+    const { getByRole } = render(<Link href="#section">Anchor Link</Link>)
+    const link = getByRole('link')
+
+    fireEvent.click(link)
+    expect(pushMock).not.toHaveBeenCalled()
+  })
+
+  it('preloads route when in viewport and preload is true', (): void => {
+    render(
+      <Link href="/test" preload={true}>
+        Test Link
+      </Link>,
+    )
+
+    intersectionObserverCallback?.(true)
+    expect(preloadMock).toHaveBeenCalled()
+  })
+
+  it('does not preload route when preload is false', (): void => {
+    render(
+      <Link href="/test" preload={false}>
+        Test Link
+      </Link>,
+    )
+
+    intersectionObserverCallback?.(true)
+    expect(preloadMock).not.toHaveBeenCalled()
+  })
+
+  it('does not call router.push when clicked with a modifier key', (): void => {
+    const { getByRole } = render(<Link href="/test">Test Link</Link>)
+    const link = getByRole('link')
+
+    fireEvent.click(link, { ctrlKey: true })
+    fireEvent.click(link, { metaKey: true })
+    fireEvent.click(link, { shiftKey: true })
+    fireEvent.click(link, { altKey: true })
+
+    expect(pushMock).not.toHaveBeenCalled()
+  })
+
+  it('calls onClick handler when clicked', (): void => {
+    const onClickMock = vi.fn()
+    const { getByRole } = render(
+      <Link href="/test" onClick={onClickMock}>
+        Test Link
+      </Link>,
+    )
+    const link = getByRole('link')
+
+    fireEvent.click(link)
+
+    expect(onClickMock).toHaveBeenCalledTimes(1)
+    expect(pushMock).toHaveBeenCalledWith('/test', { scroll: true })
+  })
+
+  it('calls onClick but does not navigate when clicked with a modifier key', (): void => {
+    const onClickMock = vi.fn()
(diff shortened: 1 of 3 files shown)


jacobhq:
To test in docs, replace tuono version in package.json with `file:../../packages/tuono` after building, then do pnpm install.


https://github.com/user-attachments/assets/293ae844-a72b-417d-aacd-e146b0cd49d8

packages/tuono-router/src/components/Link.tsx
```diff
@@ -42,14 +54,14 @@ export default function Link(
   const handleTransition: React.MouseEventHandler<HTMLAnchorElement> = (
     event,
   ) => {
-    event.preventDefault()
-    onClick?.(event)
-
-    if (href?.startsWith('#')) {
-      window.location.hash = href
```
Valerioageno, line 49: Shouldn't we keep this?

Valerioageno (commented):
Thanks for this. I think you are on the right track!
I think the time has come for adding also an integration test for the `Link` component. Too many conditions.
Do you mind adding them within this PR? Let me know if you need help

packages/tuono-router/src/components/Link.tsx
jacobhq: If we bring it up into the early return block above, we can just let it behave as a normal `a` tag, which I think is better

jacobhq:
> I think the time has come for adding also an integration test for the Link component. Too many conditions.

Agreed, I'm happy to do that!

packages/tuono-router/src/components/Link.tsx
```diff
@@ -42,14 +54,14 @@ export default function Link(
...
+    if (href?.startsWith('#') || isModifiedEvent(event)) {
+      // If the user is pressing a modifier key, we fall back to default behaviour of `a` tag
       return
     }
 
+    event.preventDefault()
+    onClick?.(event)
```
marcalexiei, line 63: If a user adds an `onClick` event handler on a `Link` component it won't be called when modifier keys are pressed or the href attribute starts with `#`.
I think that user event handler should be always called.

marcalexiei (changes requested):
Now all links now open in a new tab, with or without modifier key pressed.

This behaviour is caused by the code of `MdxLink`:

https://github.com/tuono-labs/tuono/blob/0f72506f957fe8a36effe788b4e44458ee7c2360/apps/documentation/src/components/MdxProvider/MdxLink/MdxLink.tsx#L34-L38

This issue was hidden by the `event.preventDefault` always executed on `Link` click event handler. 
The `target="_blank"` should be removed from `MdxLink` component when rendering internal links.

packages/tuono-router/src/components/Link.tsx
jacobhq: So if the user passes `onClick` do we still continue? That's what I've just done in latest commit, take a look.

marcalexiei: For the moment you implementation is fine.
In the future we might consider adding a checking on [defaultPrevented](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented) property to allow user to prevent default routing navigation.

marcalexiei (commented):
https://github.com/tuono-labs/tuono/pull/534#issuecomment-2645910250

Request a new review once test have been written, thanks.

packages/tuono-router/src/components/Link.spec.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+describe('Link Component', (): void => {
+  beforeEach((): void => {
+    pushMock.mockReset()
+    preloadMock.mockReset()
+    intersectionObserverCallback = undefined
+  })
+
+  it('renders with correct href and text', (): void => {
+    const { getByRole } = render(<Link href="/test">Test Link</Link>)
```
marcalexiei, line 40: Do not use matcher from the render method.
Use screen methods from `screen` object exported from testing library:

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

packages/tuono-router/src/components/Link.spec.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const link = getByRole('link', { name: 'Test Link' })
+
+    expect(link.getAttribute('href')).toBe('/test')
+    expect(link.textContent).toBe('Test Link')
+  })
+
+  it('calls router.push on normal click', (): void => {
```
marcalexiei, line 47: Return types are not required in test funtions (remove them in all involved functions ()).

```suggestion
  it('calls router.push on normal click', () => {
```

packages/tuono-router/src/components/Link.tsx
```diff
@@ -42,14 +54,15 @@ export default function Link(
   const handleTransition: React.MouseEventHandler<HTMLAnchorElement> = (
     event,
   ) => {
-    event.preventDefault()
     onClick?.(event)
 
-    if (href?.startsWith('#')) {
-      window.location.hash = href
+    if (href?.startsWith('#') || isModifiedEvent(event)) {
+      // If the user is pressing a modifier key, we fall back to default behaviour of `a` tag
```
marcalexiei, line 60:
```suggestion
if (
      href?.startsWith('#') ||
      // If the user is pressing a modifier key, 
      // we fall back to default behaviour of `a` tag
      isModifierKeyPressed(event)
    ) {
```

packages/tuono-router/src/components/Link.tsx
```diff
@@ -18,6 +18,18 @@ interface TuonoLinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
   scroll?: boolean
 }
 
+function isModifiedEvent(event: React.MouseEvent): boolean {
```
marcalexiei, line 21:
```suggestion
function isModifierKeyPressed(event: React.MouseEvent): boolean {
```

packages/tuono-router/src/components/Link.spec.tsx
```diff
@@ -0,0 +1,129 @@
...
+    preloadMock.mockReset()
+    intersectionObserverCallback = undefined
+  })
+
+  it('renders with correct href and text', (): void => {
+    const { getByRole } = render(<Link href="/test">Test Link</Link>)
+    const link = getByRole('link', { name: 'Test Link' })
+
+    expect(link.getAttribute('href')).toBe('/test')
+    expect(link.textContent).toBe('Test Link')
```
marcalexiei, line 44: This match is redundant with the `getByRole` parameters.
```suggestion
```

packages/tuono-router/src/components/Link.tsx
jacobhq: I think we should keep this name (or change it to something else) since we also check for the target attribute.

marcalexiei: Use `isEventModifierKeyActiveAndTargetDifferentFromSelf`

packages/tuono-router/src/components/Link.tsx
```diff
@@ -18,6 +18,18 @@ interface TuonoLinkProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
...
+  const eventTarget = event.currentTarget as HTMLAnchorElement | SVGAElement
+  const target = eventTarget.getAttribute('target')
```
marcalexiei, line 23:
```suggestion
function isEventModifierKeyActiveAndTargetDifferentFromSelf(
  event: React.MouseEvent<HTMLAnchorElement>,
): boolean {
  const target = event.currentTarget.getAttribute('target')
```

packages/tuono-router/src/components/Link.tsx
jacobhq: Ok

marcalexiei (approved):

packages/tuono-router/src/components/Link.spec.tsx
marcalexiei: #540

Issue #253, opened 2024-12-23
Add `e2e` test to tuono


Valerioageno:
We need to find a way to reliably prevent to introduce bugs on the mainstream. With this task, we are going to investigate over creating an `e2e` test.

m-jovan:
Would love to pick this up and investigate 👋

PR #535, opened 2025-02-08
feat: Add e2e setup and tests

m-jovan:
### Checklist

### Related issue

Fixes https://github.com/tuono-labs/tuono/issues/253

### Overview

In this PR we are attempting to add an end-to-end (e2e) tests in order to verify updates are not breaking the full workflow of creating a new app and running it.

These tests consist of two parts, test setup and Playwright client tests. Test setup covers building tuono, creating a new app using the tutorial template and installing dependencies. These actions are simply performed in order to mimic starting from zero and creating a new app. Possible errors in this phase are covered by integration tests in the proper manner so e2e tests are not checking anything here. Second phase is running `tuono dev` and verifying that we can access client app and do a basic action such as navigate to a route. Here, with the help of Playwright, we are checking if this action can be performed and if the client app renders.


m-jovan:
@Valerioageno @marcalexiei Sorry about force pushing and changes, I've used the PR as a draft to make sure E2E CI works and that is tweaked correctly.

marcalexiei:
> Sorry about force pushing and change

No problem, you can force push how many times you want to test CI.
Just don't don't do it after a review has been performed 😅

I marked this PR as draft, once completed you can mark it as ready for review 🙇

m-jovan:
The pipeline is green and I'm marking this as ready to review. I consider this PR as a proposal on one way we can handle e2e tests and I'm completely open to other ideas. One thing I've noticed is that the windows e2e test job is much slower than others but works.

.github/workflows/e2e-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Setup rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
+
+      - name: Install Playwright Browsers
+        run: npx playwright install
+
+      - name: Run Playwright E2E tests
+        run: npx playwright test --config=./e2e/playwright.config.ts
```
Valerioageno, line 43: We should consider adding a command in the root `package.json` like `pnpm test:e2e`

e2e/e2e-test-setup.js
Valerioageno, line 1: `tuono new` is getting properly tested on a specific integration test. With @marcalexiei we were thinking about creating a folder like `e2e/fixtures/base` and putting there a mock project to test. WDYT?

e2e/e2e-test-setup.js
```diff
@@ -0,0 +1,26 @@
+import { execSync } from 'child_process'
+import * as path from 'path'
+import { fileURLToPath } from 'url'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+
+const appDir = path.join(__dirname, 'test-app')
+
+execSync('cargo build --release', { stdio: 'inherit' })
```
Valerioageno, line 10: The `--relase` flag just improve the built binary but does not have any impact. To speed up the build process, I'd do: 

```suggestion
execSync('cargo build -C opt-level=0', { stdio: 'inherit' })
```

e2e/tests/e2e.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { test, expect } from '@playwright/test'
```
Valerioageno, line 1: e2e are very cool also because we can do http requests testing also the server side rendered HTML output!

Valerioageno (commented):
Hey @m-jovan, thanks for this huge improvement. The PR looks great. I just left a couple of comments about how structuring them, considering in the future adding more scenarios with different apps. Something like:
```
e2e/
   fixtures/
       base/ (our current case)
       scenario-with-db/
       scenario-with-something-else/
       ... so on so forth
```
For now just the base scenario works perfectly. Let me know your thoughts

.github/workflows/e2e-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Install NodeJS Dependencies
+        uses: ./.github/actions/install-node-dependencies
+
+      - name: Setup rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
+
+      - name: Install Playwright Browsers
+        run: npx playwright install
```
marcalexiei, line 40:
```suggestion
        run: pnpm exec playwright install
```

e2e/e2e-test-setup.js
```diff
@@ -0,0 +1,26 @@
+import { execSync } from 'child_process'
+import * as path from 'path'
+import { fileURLToPath } from 'url'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
```
marcalexiei, line 6: We are on node 22, maybe we can consider access `import.meta` properties to read this:
- [import.meta.filename](https://nodejs.org/api/esm.html#importmetadirname) 
- [import.meta.dirname](https://nodejs.org/api/esm.html#importmetafilename)

e2e/e2e-test-setup.js
```diff
@@ -0,0 +1,26 @@
...
+  'release',
+  isWindows ? 'tuono.exe' : 'tuono',
+)
+
+execSync(`${tuonoBuildDir} new test-app --template tuono-tutorial`, {
+  cwd: __dirname,
+  stdio: 'inherit',
+})
+
+execSync('npm install', { cwd: appDir, stdio: 'inherit' })
```
marcalexiei, line 26:
```suggestion
execSync('pnpm install', { cwd: appDir, stdio: 'inherit' })
```

e2e/playwright.config.ts
```diff
@@ -0,0 +1,29 @@
+import * as path from 'path'
+
+import { fileURLToPath } from 'url'
+
+import { defineConfig } from '@playwright/test'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
```
marcalexiei, line 8: Also here we could use `import.meta` properties

marcalexiei (commented):
Great job! 
Surely e2e tests are a marvellous addition to the project.
Thanks again for working on this 🙇

I left a few minor comments about minor code improvements.

e2e/e2e-test-setup.js
m-jovan: You mean `cargo build --config opt-level=0` ? Because `-C` flag should stand for cwd

e2e/e2e-test-setup.js
m-jovan: So to make sure I understand, we don't want to run `tuono new` and instead have these mocks already set up in `fixtures` folder? Having multiple cases completely makes sense, though my idea of e2e test was to scaffold the project from zero and run it. Of course if you have a strong opinion that we don't have to run `tuono new` we can go with hardcoded project mocks 👍

e2e/e2e-test-setup.js
m-jovan: It was intentional since here we could use any package manager so I wanted to have a default use case. Do you think `pnpm` would bring more benefits in this case?

e2e/playwright.config.ts
m-jovan: Fixed, thanks 👌

e2e/e2e-test-setup.js
m-jovan: Fixed, thanks 👌

.github/workflows/e2e-ci.yml
m-jovan: Fixed, thanks 👌

.github/workflows/e2e-ci.yml
m-jovan: Added, thanks 👌

e2e/e2e-test-setup.js
Valerioageno: Yeah, I think it makes sense to skip the `tuono new` command entirely.

e2e/e2e-test-setup.js
marcalexiei: Usually `pnpm` is faster than `npm` . 
I would consider using it because Windows workflows can be a real pain when it comes to execution time 😂.

e2e/e2e-test-setup.js
m-jovan: Added the base fixture 👌

e2e/e2e-test-setup.js
m-jovan: That makes sense, I also noticed Windows workflows are slow. Should be fixed now 👍

e2e/fixtures/base/src/routes/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <p className="subtitle">{data?.subtitle}</p>
+        <a
+          href="https://github.com/tuono-labs/tuono"
+          target="_blank"
+          className="button"
+          type="button"
+        >
+          Github
+        </a>
+      </div>
```
marcalexiei, line 45: I don't think we need this part and all associated CSS

```suggestion
```

e2e/fixtures/base/src/routes/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <>
+      <header className="header">
+        <a href="https://crates.io/crates/tuono" target="_blank">
+          Crates
+        </a>
+        <a href="https://www.npmjs.com/package/tuono" target="_blank">
+          Npm
+        </a>
+      </header>
```
marcalexiei, line 25: I don't think we need this part and all associated CSS

```suggestion
```

marcalexiei (commented):
Left a few comments about additional improvements

e2e/fixtures/base/src/routes/index.tsx
m-jovan: Removed 👍

e2e/fixtures/base/README.md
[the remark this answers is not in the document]
m-jovan: Removed 👍

e2e/fixtures/base/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "tuono-app",
+  "description": "Basic tuono application",
+  "version": "0.0.1",
+  "dependencies": {
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0",
+    "tuono": "0.17.7"
```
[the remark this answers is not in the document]
m-jovan: Indeed that makes complete sense, I've changed both to use workspace package and added a `turbo build` to the setup so we can properly use the package in when running tests.

.github/workflows/e2e-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install NodeJS Dependencies
+        uses: ./.github/actions/install-node-dependencies
+
+      - name: Setup rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
```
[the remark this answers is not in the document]
m-jovan: Removed 👍

.github/workflows/e2e-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+      - 'main'
+  pull_request:
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  e2e:
+    name: Build Tuono, create a new app and test it
```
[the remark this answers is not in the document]
m-jovan: Updated 👌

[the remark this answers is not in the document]
m-jovan: Matrix.os is already there though so I'll just remove it and leave the "E2E TEST"

`E2E CI / E2E Test ({{ matrix.os }}) (ubuntu-latest) (pull_request) `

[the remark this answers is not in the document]
marcalexiei: I forgot to add a `$` in the snippet 😅.
This is the correct snippet:
```
name: E2E Test (${{ matrix.os }})
```


GitHub Actions workflow adds the matrix value at the end if not added into the name.
I usually prefer to add it explicitly.

For example, check `rust-ci.yml`:

https://github.com/tuono-labs/tuono/blob/0114867fe63a707145876eab7a1d6d84ca6cdc43/.github/workflows/rust-ci.yml#L18

Produces

image

----

e2e/tests/e2e.spec.ts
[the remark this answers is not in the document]
Valerioageno: I wouldn't overtest. The API are tested on rust integration tests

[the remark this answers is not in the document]
marcalexiei: I added this comment as a follow up of https://github.com/tuono-labs/tuono/pull/535#discussion_r1948128673.
But no problem. Marking as resolved.

e2e/fixtures/base/src/routes/api/health_check.rs
Valerioageno, line 1: Since we are not testing this we can delete the folder

e2e/fixtures/base/public/favicon.ico
Valerioageno, line 1: I'd remove also the `public/` folder. We don't need to load images in this iteration IMHO

e2e/fixtures/base/src/routes/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  isLoading,
+}: TuonoProps<IndexProps>): JSX.Element {
+  if (isLoading) {
+    return <h1>Loading...</h1>
+  }
+
+  return (
+    <div className="title-wrap">
+      <h1 className="title">
+        TU<span>O</span>NO
```
Valerioageno: This structure can be simplified. Maybe just asserting:
1. A static string
2. A string passed from the server

e2e/fixtures/base/src/styles/global.css
```diff
@@ -0,0 +1,67 @@
+@import url('https://fonts.googleapis.com/css2?family=Poppins:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap');
```
Valerioageno, line 1: We can get rid of the CSS

> Note. the file cannot be deleted

e2e/tests/e2e.spec.ts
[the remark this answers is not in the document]
m-jovan: Agree, added 👍

e2e/e2e-test-setup.js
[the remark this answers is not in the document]
m-jovan: It is definitely more future proof so I agree. I've moved tests and playwright config to the fixture while leaving the script outside since it can be reused as it is now simplified to just prepare tuono. Let me know what you think!

.github/workflows/e2e-ci.yml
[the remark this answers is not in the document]
m-jovan: Updated

e2e/fixtures/base/src/routes/api/health_check.rs
m-jovan: Removed 👍

e2e/fixtures/base/public/favicon.ico
m-jovan: Removed 👍

e2e/fixtures/base/src/styles/global.css
m-jovan: Removed 👍

e2e/fixtures/base/src/routes/index.tsx
marcalexiei: I think point 2 is still missing. 
There is already a rust route in place: `e2e/fixtures/base/src/routes/index.rs`.
Could you please render the string in the page and check that it exists?

.gitignore
```diff
@@ -28,4 +28,7 @@ vite.config.ts.timestamp*
 *.pdb
 
 # Files, generated by the code editors
-.vscode/
\ No newline at end of file
+.vscode/
+
+/e2e/fixtures/base/node_modules/
+
```
marcalexiei, line 34: This should not be needed

```suggestion
```

e2e/fixtures/base/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "fixture-base",
+  "type": "module",
+  "description": "Basic tuono application",
+  "version": "0.0.1",
+  "scripts": {
+    "test:e2e": "pnpm exec playwright test"
```
marcalexiei, line 7: `pnpm exec` should not be needed inside package.json scripts

```suggestion
    "test:e2e": "playwright test"
```

e2e/fixtures/base/pnpm-lock.yaml
marcalexiei, line 1: Now `e2e/fixtures/base` is part of pnpm workspace,
so a `pnpm-lock.yml` file should not be present in this folder.

e2e/fixtures/base/src/routes/index.tsx
m-jovan: Updated 👌

.gitignore
m-jovan: Indeed, updated 👍

e2e/fixtures/base/package.json
m-jovan: Updated 👍

e2e/fixtures/base/pnpm-lock.yaml
m-jovan: Removed 👍

marcalexiei (approved):
Looks good to me!
Thanks again for putting your efforts into this 🙇.

Valerioageno (approved):
⚡

PR #411, opened 2025-01-24
test: add full integration test for tuono new CLI command

Robert-maker1994:
Close #405

PR #558, opened 2025-02-12
feat: init new git repo for new projects

### LINKED ISSUE #554: Automatic `git init`
Automatically init git after `tuono new` if the git command is available.

jacobhq:
### Checklist

### Related issue

Fixes #554 

### Overview

Checks if git is installed, and init a new repo by default. If git is not installed, we continue in silence, and only fail with error if the user specifies the git flag, but git is not installed.

### Tasks

- [x] Ensure git is available in CI to run tests
- [x] Write unit test for function
- [x] Write integration test for all cases
- [x] ~~Add flag to docs (not much to do, but will update the help snippet on the [cli page](https://tuono.dev/documentation/cli))~~


jacobhq:
@Valerioageno would you mind taking a look at the build on macOS? Not sure why it's failing - I think it may not be related to this PR.

marcalexiei:
On #560 Rust CI runs without error on MacOS:
 https://github.com/tuono-labs/tuono/actions/runs/13293248776/job/37118888545?pr=560

marcalexiei:
Rerunning the failed job hasn't produced any error: 
https://github.com/tuono-labs/tuono/actions/runs/13291663120/job/37119315943?pr=558

jacobhq:
> Rerunning the failed job hasn't produced any error: https://github.com/tuono-labs/tuono/actions/runs/13291663120/job/37119315943?pr=558

Thank you!

marcalexiei (commented):
I added a few comments to refine messages and comments.
Let me know what do you think 😉

(The code review will still be performed by @Valerioageno  since I'm not a rust expert 😅)

jacobhq:
> I added a few comments to refine messages and comments.

Thank you! What do you think of my proposed logic for the git workflow (description in PR overview)?

crates/tuono/src/cli.rs
```diff
@@ -36,6 +36,9 @@ enum Actions {
         /// Load the latest commit available on the main branch
         #[arg(long)]
         head: Option<bool>,
+        /// Initialize a new Git repository in the specified directory
+        #[arg(short, long)]
+        git: Option<bool>,
```
marcalexiei, line 41: > What do you think of my proposed logic for the git workflow?

It works for me: by default, `git init` is executed, and an error is thrown if Git is not installed. 
The user can override this behavior by setting the flag to `false`.

However I would consider to use a more somatic name for the flag. E.g.: `--git-init` rather than just `git`.

crates/tuono/tests/cli_build.rs
marcalexiei, line 1: Anyway `cargo test` is failing again because the command output is:
```text
Failed to parse the tag response
```
rather than exiting with a success status code.
This makes me wondering if there is a temporary issue with GitHub API since rerunning produced, again, no errors.

@Valerioageno I think this should be resolved by using a mock server when executing unit tests,
which is already worked on #411.
However the work is halted due to a missing release of `.dotenvy` crate. 
Maybe there is another way to go ahead? 
E.g. Creating a trait hardcoding the GitHub domain for production and mocking it with the local mock server during unit tests?

crates/tuono/src/cli.rs
Valerioageno: > and an error is thrown if Git is not installed.

To me, no error should be thrown. The only error should happen when the user specifically use `--git-init` and git is not installed

crates/tuono/src/scaffold_project.rs
```diff
@@ -66,9 +67,18 @@ pub fn create_new_project(
...
 ) {
     let folder = folder_name.unwrap_or(".".to_string());
 
+    // Check if Git is installed when the user requests to use it; otherwise, continue
+    if git.unwrap_or(false) && !is_git_installed() {
+        exit_with_error("You requested to use Git, but it is not installed.")
+    }
+
+    // Use git by default
+    let git = git.unwrap_or(true) && is_git_installed();
```
Valerioageno, line 80: I'd avoid running the command twice

```suggestion
    let is_git_installed = is_git_installed();

    // Check if Git is installed when the user requests to use it; otherwise, continue
    if git.unwrap_or(false) && !is_git_installed {
        exit_with_error("You requested to use Git, but it is not installed.")
    }

    // Use git by default
    let git = git.unwrap_or(true) && is_git_installed;
```

crates/tuono/src/scaffold_project.rs
```diff
@@ -151,6 +161,11 @@ pub fn create_new_project(
 
     update_package_json_version(&folder_path).expect("Failed to update package.json version");
     update_cargo_toml_version(&folder_path).expect("Failed to update Cargo.toml version");
+
+    if git {
+        init_new_git_repo(&folder_path).expect("Failed to initialise a new git repo");
```
Valerioageno, line 166: We should never make the command panic

```suggestion
        init_new_git_repo(&folder_path).unwrap_or_else(|| exit_with_error("Failed to initialise a new git repo"));
```

crates/tuono/src/scaffold_project.rs
```diff
@@ -292,4 +329,25 @@ mod tests {
         );
         assert_eq!(expected, generated)
     }
+
+    #[test]
+    fn test_init_new_git_repo() {
```
Valerioageno, line 334: I'd remove this test. The mock is not good enough since the success/fail of the test depends on the running machine (If `git` is not installed it fails)

crates/tuono/tests/cli_build.rs
Valerioageno: Yeah, definitely needs to create the mock first.

Let's put these tests in a separate branch and merging them once the above PR gets merged.

Note: These tests should be put in `tests/cli_new.rs`

crates/tuono/tests/cli_build.rs
```diff
@@ -200,3 +200,72 @@ fn build_fails_with_no_config() {
         .failure()
         .stderr("Cannot find tuono.config.ts - is this a tuono project?\n");
 }
+
+#[test]
+#[serial]
+fn it_inits_new_git_repo_by_default_with_git_installed() {
+    let temp_tuono_project = TempTuonoProject::new();
```
Valerioageno, line 207: The same problem in the unit test happens here (if the host machine has not installed `git` the tests fail).
We should consider mocking it, otherwise we gonna increase test complexity and flakiness

crates/tuono/src/cli.rs
jacobhq: > To me, no error should be thrown. The only error should happen when the user specifically use --git-init and git is not installed

That's what happens currently yep.

I will rename the flag to `--git-init` though.

jacobhq:
Please can I have a bit of help here? When I use `TempTuonoProject` in the `cli_build.rs` file, there are no errors. However, because the rust compiler treats these tests separately, when I use the module in the `cli_new.rs` file, the compiler errors because in this test, the function is not used.

I could ignore the dead code warning, but that feels bad. WDYT?

jacobhq:
@Valerioageno should I just remove some of these tests or comment them out/disable them so they can be rewritten once we have a mock?

Valerioageno:
Right now the tests are not reliable enough to be merged since they strongly rely on the host machine that run them (in a machine that doesn't have installed `git` they just fail).

I think that for this iteration we could:
1. merge this PR without tests
2. Complete the other PR (let me know if you are interested in taking it over)
3. Add tests for this scenario

WDYT?

jacobhq:
Should we put this on hold until we get the mock (I'm happy to help with that if needed), and then merge those changes into this branch and refactor the tests as needed?

I think it's a bit risky to merge this one without tests.

Valerioageno:
Makes sense! The goal with the other PR is to find a library that supports configuration at build time. When we have that we can proceed with this then! :)

Valerioageno:
This PR is unblocked now!

jacobhq:
So to be clear, the only thing left to do is write some kind of mock for git?

Valerioageno:
Exactly, otherwise the test is dependant on the software installed on the running machine.

Valerioageno:
Let me know @jacobhq if you need support with this! :)

Valerioageno (commented):
I'd easier down it even more.

Let's remove the `--git-init` flag (all the installations also init git if existing).
This way we can also skip the check `is_git_installed` and simply run the command at the end of the routine.

Let's then remove the error handling in `init_new_git_repo` and simply check trace every error with a meaningful message.

WDYT?

crates/tuono/src/commands/new.rs
```diff
@@ -265,6 +271,22 @@ fn update_cargo_toml_version(folder_path: &Path) -> io::Result<()> {
...
+    } else {
+        Err(io::Error::new(
+            io::ErrorKind::Other,
+            format!(
+                "Git init failed: {}",
+                String::from_utf8_lossy(&output.stderr)
+            ),
+        ))
+    }
+}
```
Valerioageno, line 288: I don't see the point of returning a `Result` for logging a generic trace. Couldn't we manage a more granular logging?

```suggestion
fn init_new_git_repo(folder_path: &Path) {
    if let Ok(output) = Command::new("git").arg("init").arg(folder_path).output() {
       if output.status.is_err() {
          trace!(...)
       }
    } else {
        trace!(...)
    }
}
```

Valerioageno (approved):

PR #560, opened 2025-02-12
feat(crates/tuono): feat: add spinner to `tuono build`

### LINKED ISSUE #285: [Feature request]: `tuono dev`/`tuono build` seems kind of slow

At the moment, `tuono dev` and `tuono build` seem sluggish. Rather than actual performance issues, I think this issue boils down to`tuono dev` and `tuono build` not really showing any output in the terminal, making the app look slow

I think that a cargo-like output would be a clean way to implement logging and responsiveness within tuono.

Example:
```
   Compiling hello-world v0.1.0 (/tmp/tmp.AdsdVoKBky/hello-world)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/hello-world`
```

Something like this could work for `tuono dev`:
```
  Building   Frontend
  Finished   React Frontend
  Compiling  Rust Backend
    < Output of `cargo build` or whatever is building the backend >
  Serving    at localhost:8080
```

The example above for `tuono dev` is a really quick draft of what `tuono build` and `tuono dev` would look like once this issue has been fixed.

_No response_

### Additional context

_No response_

marcalexiei:
### Checklist

### Related issue

Fixes #285

### Overview

Show a spinner during `tuono build` process.
Moved build in a separate rust module to be consistent with watch module.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,3 +5,6 @@ end_of_line = lf
 indent_style = space
 indent_size = 2
 max_line_length = 80
+
+[*.rs]
+indent_size = 4
--- a/crates/tuono/src/build.rs
+++ b/crates/tuono/src/build.rs
@@ -1,7 +1,5 @@
-use console::Term;
 use fs_extra::dir::{copy, CopyOptions};
 use spinners::{Spinner, Spinners};
-use std::io;
 use std::path::PathBuf;
 use std::thread::sleep;
 use std::time::Duration;
@@ -9,32 +7,33 @@ use std::time::Duration;
 use crate::app::App;
 use crate::mode::Mode;
 
-#[tokio::main]
-pub async fn build(mut app: App, ssg: bool, no_js_emit: bool) -> Result<(), io::Error> {
+pub fn build(mut app: App, ssg: bool, no_js_emit: bool) {
+    let mut app_build_spinner = Spinner::new(Spinners::Dots, "Building app...".into());
+
     if no_js_emit {
         println!("Rust build successfully finished");
-        return Ok(());
+        return;
     }
 
     if ssg && app.has_dynamic_routes() {
         // TODO: allow dynamic routes static generation
         println!("Cannot statically build dynamic routes");
-        return Ok(());
+        return;
     }
 
     app.build_tuono_config()
         .expect("Failed to build tuono.config.ts");
 
-    // start showing the spinner
-    let term = Term::stdout();
-    let mut sp = Spinner::new(Spinners::Dots, "Building app...".into());
-
     app.check_server_availability(Mode::Prod);
 
     app.build_react_prod();
 
+    // Remove the spinner
+    app_build_spinner.stop_with_message("\u{2705}Build completed".into());
+
     if ssg {
-        println!("SSG: generation started");
+        let mut app_build_static_spinner =
+            Spinner::new(Spinners::Dots, "Static site generation".into());
 
         let static_dir = PathBuf::from("out/static");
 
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -5,3 +5,6 @@ end_of_line = lf
 indent_style = space
 indent_size = 2
 max_line_length = 80
+
+[*.rs]
+indent_size = 4
```
marcalexiei, line 10: Unrelated to this PR... but very useful in the code editor 😅

crates/tuono/src/build.rs
```diff
@@ -0,0 +1,89 @@
...
+use std::io;
+use std::path::PathBuf;
+use std::thread::sleep;
+use std::time::Duration;
+
+use crate::app::App;
+use crate::mode::Mode;
+
+#[tokio::main]
+pub async fn build(mut app: App, ssg: bool, no_js_emit: bool) -> Result<(), io::Error> {
```
Valerioageno, line 13: No need for `tokio::main`, `async` and a result type
```suggestion
pub fn build(mut app: App, ssg: bool, no_js_emit: bool) {
```

crates/tuono/src/build.rs
```diff
@@ -0,0 +1,89 @@
...
+        // Close server
+        let _ = rust_server.kill();
+    }
+
+    // Remove the spinner
+    sp.stop();
+    term.clear_line().unwrap();
+
+    println!("Build successfully finished");
+    Ok(())
```
Valerioageno, line 88:
```suggestion
```

crates/tuono/src/build.rs
```diff
@@ -0,0 +1,89 @@
...
+        // TODO: allow dynamic routes static generation
+        println!("Cannot statically build dynamic routes");
+        return Ok(());
+    }
+
+    app.build_tuono_config()
+        .expect("Failed to build tuono.config.ts");
+
+    // start showing the spinner
+    let term = Term::stdout();
```
Valerioageno, line 29: I'd remove the Term interface to replace the spinner with a ✅ rather than removing the line below
```suggestion
```

crates/tuono/src/build.rs
```diff
@@ -0,0 +1,89 @@
...
+            route.save_ssg_file(&reqwest_client);
+        }
+
+        // Close server
+        let _ = rust_server.kill();
+    }
+
+    // Remove the spinner
+    sp.stop();
+    term.clear_line().unwrap();
```
Valerioageno, line 85:
```suggestion
    sp.stop_with_symbol("\x1b[32m🗸\x1b[0m");
```

crates/tuono/src/cli.rs
```diff
@@ -77,78 +75,9 @@ pub fn app() -> std::io::Result<()> {
...
-                for (_, route) in app.route_map {
-                    route.save_ssg_file(&reqwest)
-                }
-
-                // Close server
-                let _ = rust_server.kill();
-            };
-
-            println!("Build successfully finished");
+            let _ = build::build(app, ssg, no_js_emit);
```
Valerioageno, line 80:
```suggestion
            build::build(app, ssg, no_js_emit)
```

crates/tuono/src/build.rs
Valerioageno: Also I'd put this at the beginning of the function

Valerioageno (commented):
I'd do two spinners:
1. The basic build 
2. The SSG build

The output should be something like:

```
🗸 Building app...
🗸 Static site generation
```

crates/tuono/src/build.rs
marcalexiei: This is causing test fail because the spinner text is now included in the `stderr` which is checked by the following tests:
- `it_fails_without_installed_build_config_script`
- `it_fails_without_installed_build_script`

These tests check errors that are coming from `App` `build_tuono_config` function which exit the process with an error status, so there is no way to remove, nor replace the spinner from the `build` function.

IMHO we should return these errors to the caller,
so they can be properly managed, e.g.: display custom error messages / hiding spinner / exit the process.

However this seems out of scope for this PR, so I suggest we simply move `app_build_spinner` declaration `app.check_server_availability(Mode::Prod);` and treat this as a separate task.

Let me know what do you think about this.

marcalexiei: As per Discord discussion we are going ahead using the workaround.

PR #569, opened 2025-02-15
refactor: apply RustRover lint suggestions

jacobhq:
### Checklist

### Related issue

Fixes None

### Overview

As discussed, apply suggestions from RustRover. I agree about not doing the `Default` implementation in `utils.rs`.

- [x] What do you think about a `mod.rs` and `utils.rs` being present in `crates/tuono_lib/tests/utils` (we end up with utils nested twice). We could rename, or put the contents of `utils.rs` into `mod.rs` (I think that is best option). https://rust-lang.github.io/rust-clippy/master/index.html#module_inception

For reference, these were the raised (weak) warnings:

```
file mod.rs   1 weak warning 
ERROR Module has the same name as its containing module 
file payload.rs   1 weak warning 
ERROR The following explicit lifetimes could be elided: 'a 
file server_test.rs   4 weak warnings 
ERROR The borrowed expression implements the required traits 
ERROR The borrowed expression implements the required traits 
ERROR The borrowed expression implements the required traits 
ERROR The borrowed expression implements the required traits 
file utils.rs   1 weak warning 
ERROR You should consider adding a `Default` implementation for `TempTuonoProject` 
file utils.rs   1 weak warning 
ERROR Unnecessary use of `to_owned` 
file utils.rs   3 weak warnings 
ERROR Use of `expect` followed by a function call 
ERROR Non‑binding `let` on a future 
ERROR Unnecessary use of `to_owned` 
```

So to conclude, in this PR, I want to fix all except "adding a `Default` implementation for `TempTuonoProject`"

Please let me know your thoughts!


examples/tuono-tutorial/src/app.rs
```diff
@@ -7,5 +7,5 @@ pub struct ApplicationState {
 
 pub fn main() -> ApplicationState {
     let fetch = Client::new();
-    return ApplicationState { fetch };
+    ApplicationState { fetch }
```
Valerioageno, line 10: Let's not update this for now. We should also update the documentation

crates/tuono_lib/tests/utils/utils.rs
```diff
@@ -86,7 +85,7 @@ impl MockTuonoServer {
             .local_addr()
             .expect("Failed to extract test server socket");
 
-        let _ = tokio::spawn(server.start());
+        _ = tokio::spawn(server.start());
```
Valerioageno, line 88: TIL!

jacobhq: That syntax is a bit weird, and I didn't know about it before either! It just skips binding the result of `tokio::spawn(server.start())` to variable and instead drops it immediately, as opposed to binding and then dropping. It's equivalent to calling `std::mem::drop(tokio::spawn(server.start()));`, which is more explicit.

examples/tuono-tutorial/src/app.rs
jacobhq: Should I revert this back to being a `return`, or should I keep it and update the docs to reflect it?

Valerioageno: I'd revert for now. Just less work

crates/tuono_lib/tests/utils/utils.rs
jacobhq, line 1: This file is nested in the utils module, so we have to import with `utils::utils`. We could move this into mod.rs to prevent this. WDYT?

Valerioageno: Agree with the mod!

jacobhq (commented):
Alright, ready to go

crates/tuono_lib/tests/utils/mod.rs
```diff
@@ -1,3 +1,108 @@
+use fs_extra::dir::create_all;
```
Valerioageno, line 1: `mod.rs` file should be just a tunnel for the files. Could you please put the implementation in a separate file?

PR #583, opened 2025-02-21
feat(env): add environment variables with hot reload for rust code

### LINKED ISSUE #570: Enhancement: Implement .env file support for configuration management

### LINKED ISSUE #586: Add env manager in tuono CLI

Refer to -> https://github.com/tuono-labs/tuono/issues/570

Refer to -> https://github.com/tuono-labs/tuono/issues/570

_No response_

### Additional context

_No response_

jacobhq:
### Checklist

### Related issue

Fixes #586 

### Overview

Rust EnvManager struct and hot reload for serverside env changes (and tests).

New env files created after server start are not loaded in, however any changes to the env files already loaded are auto-applied. This behavior is consistent with vite.


crates/tuono/src/app.rs
```diff
@@ -199,6 +203,7 @@ impl App {
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
+            .envs(std::env::vars())
```
Valerioageno, line 206: I'm not a fan of having the vars in both processes (CLI and server).
What about passing the var using a vector?

```suggestion
            .envs(self.env_var_manager.variables)
```

Valerioageno (commented):
Left few comments.

If I understand correctly we add the variables to the current process and then we start the server cloning all the variables.

if so I'd suggest to just pass the variables to the server without this back and forth.

Also I noticed that we have a split between `system_vars` (I think are the variables set with the `tuono` command [do we need to support them?]) and the defined in the `.env` files. Do we need such split?

---

For the record, I'd personally split the `env_manager` and `watch reload` PRs. Too broad the scopes

crates/tuono/src/app.rs
```diff
@@ -54,13 +55,16 @@ fn has_app_state(base_path: PathBuf) -> std::io::Result<bool> {
 
 impl App {
     pub fn new() -> Self {
+        let env_var_manager = EnvVarManager::new(Mode::Prod);
```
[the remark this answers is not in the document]
jacobhq: The `App` struct is only used in production. When it was Option<Mode> it was `None`, but since we only use `App` in production, I think this is fine. WDYT?

crates/tuono/src/env.rs
```diff
@@ -0,0 +1,206 @@
...
+        env_files.push(format!(".env.{}", mode_name));
+        env_files.push(format!(".env.{}.local", mode_name));
+
+        // Collect system environment variable keys into a HashSet for fast lookup.
+        let system_vars: HashSet<String> = env::vars().map(|(k, _)| k).collect();
+
+        Self {
+            env_files,
+            system_env: system_vars,
+        }
```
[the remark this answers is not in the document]
jacobhq: Yeah I think that might be a good idea. I called it `system_vars` because initially it loads in the system environment variables, and then adds any set by the user (this was to ensure that the user can never override a system env var in an env file). I will rename as suggested :)

crates/tuono/src/env.rs
```diff
@@ -0,0 +1,206 @@
+use crate::mode::Mode;
+use std::collections::HashSet;
+use std::env;
+use std::fs;
+
+#[derive(Clone, Debug)]
+pub struct EnvVarManager {
+    env_files: Vec<String>,
+    system_env: HashSet<String>,
+}
```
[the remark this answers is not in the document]
jacobhq: Perhaps. We only scan for env files in root of project, so is `PathBuf` overkill for that?

crates/tuono/src/env.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub fn new(mode: Mode) -> Self {
+        let mut env_files = vec![String::from(".env"), String::from(".env.local")];
+
+        let mode_name = match mode {
+            Mode::Dev => "development",
+            Mode::Prod => "production",
+        };
+
+        env_files.push(format!(".env.{}", mode_name));
+        env_files.push(format!(".env.{}.local", mode_name));
```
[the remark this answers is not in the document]
jacobhq: No. I don't know what the use case is, but `.env.production.local` is valid, and should take priority over `.env.production`.

crates/tuono/src/app.rs
jacobhq: Much better idea, will do that, thanks!

jacobhq:
> If I understand correctly we add the variables to the current process and then we start the server cloning all the variables.
> 
> if so I'd suggest to just pass the variables to the server without this back and forth.

Yes that is a better idea.

> Also I noticed that we have a split between system_vars (I think are the variables set with the tuono command [do we need to support them?]) and the defined in the .env files. Do we need such split?

Yes we should, because these are also other system variables that may be present. They [should take priority](https://vite.dev/guide/env-and-mode.html#env-files:~:text=In%20addition%2C%20environment%20variables%20that%20already%20exist%20when%20Vite%20is%20executed%20have%20the%20highest%20priority%20and%20will%20not%20be%20overwritten%20by%20.env%20files.%20For%20example%2C%20when%20running%20VITE_SOME_KEY%3D123%20vite%20build.).

> For the record, I'd personally split the env_manager and watch reload PRs. Too broad the scopes

I see how that might be more aligned with atomic PRs, but then we are adding a struct with no usages. The actual bit that handles hot reload is only 10 lines or so, and whilst ordinarily I think they should be split, I think in this case it is better together since the reviewer needs to have the `EnvVarManager` in mind.

crates/tuono/src/env.rs
```diff
@@ -0,0 +1,206 @@
...
+        // Collect system environment variable keys into a HashSet for fast lookup.
+        let system_vars: HashSet<String> = env::vars().map(|(k, _)| k).collect();
+
+        Self {
+            env_files,
+            system_env: system_vars,
+        }
+    }
+
+    pub fn refresh_env_files(&self) {
```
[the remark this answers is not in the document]
jacobhq: What about `reload_variables`. It's tricky because this function initializes and refreshes.

crates/tuono/src/build.rs
```diff
@@ -32,6 +32,8 @@ pub fn build(mut app: App, ssg: bool, no_js_emit: bool) {
 
     app.check_server_availability(Mode::Prod);
 
+    app.env_var_manager.refresh_env_files();
```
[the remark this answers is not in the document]
jacobhq: I like that we explicitly call this at a specific time. IMHO it makes it easier to think about as it is quite implicit to do this in `new`. Maybe with renaming this would be better?

crates/tuono/src/env.rs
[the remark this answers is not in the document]
jacobhq:
```
let system_vars: HashSet<String> = env::vars().map(|(k, _)| k).collect();
```

Keeping that bit as is though because it is system vars. Renaming struct field.

crates/tuono/src/app.rs
jacobhq: That hash set `self.env_var_manager.variables` only contains the names. We store the actual values in `std::env`. Should we keep doing that? WDYT

Valerioageno:
> Yes we should, because these are also other system variables that may be present. They [should take priority](https://vite.dev/guide/env-and-mode.html#env-files:~:text=In%20addition%2C%20environment%20variables%20that%20already%20exist%20when%20Vite%20is%20executed%20have%20the%20highest%20priority%20and%20will%20not%20be%20overwritten%20by%20.env%20files.%20For%20example%2C%20when%20running%20VITE_SOME_KEY%3D123%20vite%20build.).

My point is that we should prevent the users to start `tuono` with some ENV variable declared from the command line. But I see your point. Better to have both. 

> I see how that might be more aligned with atomic PRs, but then we are adding a struct with no usages.

Good for me keeping everything here. I think this is not the meaning of atomic PR tho. For my experience doing atomic PR means scoping even smaller PR for every change (i.e. create struct, read var, reload var, pass var to process, etc).

This because at each review we know exactly what part of the feature you are addressing and it is easier reviewing it and challenging the implementation. 250 new code rows are very though to challenge since there are too moving parts. 

Knowing the context of a change should come from a meaningful PR description rather than the whole code (we expect that the previous code is correct since already reviewed).

No worries here. Let's keep in mind that Atomic PR is meant for better challenging the PRs to deliver higher quality

crates/tuono/src/app.rs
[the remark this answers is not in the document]
Valerioageno: Arent't we creating also app for `tuono dev`?

crates/tuono/src/build.rs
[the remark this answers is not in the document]
Valerioageno: I don't have a strong opinion. I just think that the more we encapsulate the task responsibility, the better. Otherwise for debugging we have to do more back and forth

crates/tuono/src/env.rs
```diff
@@ -0,0 +1,206 @@
...
+
+#[derive(Clone, Debug)]
+pub struct EnvVarManager {
+    env_files: Vec<String>,
+    system_env: HashSet<String>,
+}
+
+impl EnvVarManager {
+    pub fn new(mode: Mode) -> Self {
+        let mut env_files = vec![String::from(".env"), String::from(".env.local")];
```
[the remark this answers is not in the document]
Valerioageno: Agree.

Valerioageno (commented):
Alright. I'd just change the way we pass the data from the CLI to the server process (avoiding having the env cloned on both processes).

I think it is better for a number of reasons, but mostly because it is more readable, and it is not the best using the operating system capabilities for passing the data across the two processes.

crates/tuono/src/app.rs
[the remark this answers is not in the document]
jacobhq: I don't think so, but I see your point. We may do this in future, so should I add a mode param to App::new, and pass it down?

jacobhq:
> This because at each review we know exactly what part of the feature you are addressing and it is easier reviewing it and challenging the implementation. 250 new code rows are very though to challenge since there are too moving parts.
> 
> Knowing the context of a change should come from a meaningful PR description rather than the whole code (we expect that the previous code is correct since already reviewed).
> 
> No worries here. Let's keep in mind that Atomic PR is meant for better challenging the PRs to deliver higher quality

Thanks for the feedback, that has made it much clearer. In future I will try to do this.

> Alright. I'd just change the way we pass the data from the CLI to the server process (avoiding having the env cloned on both processes).

Ok. Here's my plan:

- We load all system env into a Hashmap (`env_vars` or whatever it's called) during App::new
- We load in all env values during App::new (even if that's just calling reload_variables) like you suggested. I agree - I think doing that is better for the code base overall
- We refresh this like we are doing
- We pass this hashmap as env like you suggested

My concerns with this approach:
- Loading all env is bad for performance - I don't know how to solve this since we can't differentiate between what is passed before the command and what is always present.

I've resolved all of code review, because I'd like to discuss this with you more, so that we can figure out the best way together.

I'm really open to suggestions or big changes - please let me know :)

crates/tuono/src/app.rs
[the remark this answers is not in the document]
Valerioageno: Why do you think that `App` is used only in prod? Probably (for sure) we should improve the data handling within the CLI but `App` is a graph that represents the current tuono app. The main usage is in `bundle_axum_source` that creates the `.tuono/main.rs` file (which is the axum fs routing). 

If we force prod when the users do `tuono dev` they will automatically load `.env.production` isn't it?

Valerioageno:
> Ok. Here's my plan:

Agree. (Do we need two separate variables for storing system and user defined variables?)

> Loading all env is bad for performance

Why you think that? What performance you think we are dropping?

The CLI is just a fancy interface to deal with the `tuono` project. IMHO, we should always prefer a better codebase rather a top performance process (just for the CLI).

`tuono_lib` is the server, so in this crate we have to strictly prefer improving performance rather than a more readable/understandable code.

jacobhq:
> Why you think that? What performance you think we are dropping?

We would have to load the name and value every variable present on the system, even those completely unrelated to tuono. (Run `printenv` on linux and you can see these). This is because I don't think there is a way to tell the difference betwen variables set before the command (eg `MY_VAR=hello tuono dev`), and the rest of system environment variables.

Maybe this is just not really an issue, but it doesn't feel like the *right* way to do this. In fact it looks like [this is what nodejs does anyway](https://nodejs.org/api/process.html#processenv) (and I think [vite just uses node `process.env`](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/env.ts)), so maybe let's just do it? WDYT

> The CLI is just a fancy interface to deal with the tuono project. IMHO, we should always prefer a better codebase rather a top performance process (just for the CLI).

Agree.

Valerioageno:
> This is because I don't think there is a way to tell the difference betwen variables set before the command (eg MY_VAR=hello tuono dev), and the rest of system environment variables.

Do we have to do this split? What is the benefit of knowing where they came from?

> Maybe this is just not really an issue, but it doesn't feel like the right way to do this. In fact it looks like [this is what nodejs does anyway](https://nodejs.org/api/process.html#processenv) (and I think [vite just uses node process.env](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/env.ts)), so maybe let's just do it? WDYT

I haven't understood what you are referring to... what should we do?

jacobhq:
The user's OS also has environment variables (eg `PATH`), and there are many which aren't relevant to tuono. When you set a variable like this `MY_VAR=hello tuono dev`, you have no way to distinguish it from all the other environment variables that are set on the OS. As consequence, you must load all of them into the CLI, and pass them to the server.

I was worried because I thought loading all of them was bad, however I think that is just the way they work. So I think it is the right decision to do that (https://github.com/tuono-labs/tuono/pull/583#issuecomment-2676381810), since every process should have access to all environment variables.

I hope that made a bit more sense (my worries were pretty unfounded I think, so that might be what confused you).

I will start implementing this today.

Valerioageno:
I think that is not our problem. We should just ignore them.

`tuono` crate has full access to them as well as `tuono_lib`. What we should do is translating the `.env` defined variables to `tuono_lib`. Since they are system defined, there is nothing to do with them. They will be available out of the box on any process. Isn't it?

jacobhq:
Agree. The only reason we have to interact with them to start with is so that they take precedence over ones from the file. We do still have to collect them because if we do `.envs(var_hashmap)`, it will give the sub process only the vars from that hashmap, and not the system ones. We can do `.envs(var_hashmap)` if we put the system ones into the hashmap, which is what we will do.

So the plan can stay the same:

- We load all system env into a Hashmap (env_vars or whatever it's called) during App::new
- We load in all env values during App::new (even if that's just calling reload_variables) like you suggested. I agree - I think doing that is better for the code base overall
- We refresh this like we are doing
- We pass this hashmap as env like you suggested

Valerioageno:
> We do still have to collect them because if we do .envs(var_hashmap), it will give the sub process only the vars from that hashmap, and not the system ones

Are you sure about that? I think every process has access to all the system env variables

jacobhq:
Yes you're right, I'm sorry about all that - I was wrong. I read the docs and it looks like [`.env()`](https://doc.rust-lang.org/std/process/struct.Command.html#method.env) adds to or updates them, however this takes precedence over system. I've changed the plan to prevent this and stay consistent with vite.

Updated plan then (changes are in ***bold italic***):

- We load all ***names of keys in*** system env into a Hashmap (env_vars or whatever it's called) during App::new
- We load in ***only env vales that do not exist in system*** during App::new (even if that's just calling reload_variables) like you suggested.
- We refresh this like we are doing
- We pass this hashmap as env

jacobhq:
Yes you're right, I'm sorry about all that - I was wrong. I read the docs and it looks like [`.env()`](https://doc.rust-lang.org/std/process/struct.Command.html#method.env) adds to or updates them, however this takes precedence over system. I've changed the plan to prevent this and stay consistent with vite.

Updated plan then (changes are in ***bold italic***):

- We load all ***names of keys in*** system env into a Hashmap (env_vars or whatever it's called) during App::new
- We load in ***only env vales that do not exist in system*** during App::new (even if that's just calling reload_variables) like you suggested.
- We refresh this like we are doing
- We pass this hashmap as env

Valerioageno:
We are on the same page! ⚡

jacobhq:
I think we should go back to using std::env more. I feel I am reinventing the wheel a bit here, and the code is getting incredibly complex.

Valerioageno:
> I think we should go back to using std::env more. 

What do you mean?

> I feel I am reinventing the wheel a bit here,

What are you reinventing?

crates/tuono/src/watch.rs
```diff
@@ -79,13 +81,26 @@ pub async fn watch() -> Result<()> {
     let term = Term::stdout();
     let mut sp = Spinner::new(Spinners::Dots, "Starting dev server...".into());
 
+    // Initialize EnvVarManager with Dev mode
+    let env_var_manager = Arc::new(RwLock::new(EnvVarManager::new(Mode::Dev)));
```
jacobhq, line 85: All of this feels a bit too much. It may just be me doing it wrong, but it was a lot simpler just sending env_vars to std::env, which has already been made to do this. WDYT?

Valerioageno: I don't like the idea of using a OS data structure to pass data between two processes. I prefer handling everything on the cli and passing it directly to the server process.
If it is getting too complicated or blurry I suggest to close this PR and doing a new PoC.

jacobhq:
The issue is there's not a way to pass in environment variables here: https://github.com/tuono-labs/tuono/blob/19207d7d907fc4598508a5f06f9cfb3f7c29ffb6/crates/tuono/src/watch.rs#L42C1-L62C2.

We have to load into OS env at some point - if this is scoped to process, we should move all env code out of the CLI, and into `tuono_lib`. We should then change the build and dev (ran by cargo run called in the snippet I linked) to update and include these variables, by moving `EnvManager` there. WDYT?

Valerioageno:
Yeah I think the `EnvManager` is still the right choice. 

Consider doing a poc with your proposal!

crates/tuono_lib/src/server.rs
```diff
@@ -70,12 +74,15 @@ impl Server {
...
                 .await
                 .expect("[SERVER] Failed to bind to address"),
+            env_var_manager,
         }
     }
 
     pub async fn start(self) {
         self.display_start_message();
 
+        self.env_var_manager.load_into_env();
```
Valerioageno, line 84: Why loading the variables on start? 
Couldn't we load them at `EnvManager::new()` and avoiding to add it to the struct?

crates/tuono_lib/src/env.rs
```diff
@@ -0,0 +1,227 @@
...
+            env_files,
+            system_env_names,
+            env_vars,
+        };
+
+        manager.reload_variables(); // Load only missing env variables
+        manager
+    }
+
+    pub fn reload_variables(&mut self) {
```
Valerioageno, line 39: In my opinion this should be private and be part of the `new` method

crates/tuono_lib/src/env.rs
```diff
@@ -0,0 +1,227 @@
...
+mod tests {
+    use super::*;
+    use crate::mode::Mode;
+    use serial_test::serial;
+
+    fn setup_env_file(file_name: &str, contents: &str) {
+        fs::write(file_name, contents).expect("Failed to write test .env file");
+    }
+
+    fn cleanup_env_files() {
```
Valerioageno, line 79: I think it is not rust idiomatic having functions to call at the end of each test.

The best IMHO is:

```rs
struct MockEnv;

impl Drop for MockEnv {
   fn drop() {
        // Cleanup
   }
}

#[test]
fn test_system_env_var_precedence() {
    let env = MockEnv::new();
    
    env.set_system_var(); //custom method
    env.create_env_file(); // custom method
    
    // At the end of the test it gets automatically cleaned up.
}
```

Valerioageno (commented):
Overall the implementation looks very good!
I'd add also an integration test just to be sure that the whole flow is covered.

crates/tuono_lib/src/server.rs
jacobhq: My understanding is this is what's called when the server is restarted (due to file change), so we must refresh them here. Is that correct?

crates/tuono_lib/src/env.rs
jacobhq: Wow, that is a very smart way of doing it! I will implement it :)

crates/tuono_lib/src/server.rs
Valerioageno: When we restart the server in watch the whole process restarts

crates/tuono_lib/src/lib.rs
```diff
@@ -16,6 +17,7 @@ mod ssr;
 mod vite_reverse_proxy;
 mod vite_websocket_proxy;
 
+pub use env::EnvVarManager;
```
Valerioageno, line 20: We should not expose the `EnvManager` to the userland
```suggestion
use env::EnvVarManager;
```

crates/tuono_lib/src/env.rs
```diff
@@ -0,0 +1,239 @@
...
+            env::remove_var("MISSING_EQUALS_SIGN");
+        }
+    }
+
+    #[test]
+    #[serial]
+    fn test_system_env_var_precedence() {
+        let env = MockEnv::new();
+
+        env::set_var("TEST_KEY", "system_value");
```
Valerioageno, line 113: In my opinion, we should have a `env.add_system_var` that keeps track of the added variables so that the cleanup just removes the added vars (same for the files)

crates/tuono_lib/tests/env.rs
```diff
@@ -0,0 +1,56 @@
+use std::env;
```
Valerioageno, line 1: I'd remove this file.

Integration tests should test the whole flow (not just a part of it). These tests, as they are right now, have basically the same purpose of the units.

We should just add a route in the mock server that returns in the response the value of an env variable that is sets in an env file (created in the `MockTuonoServer` struct).

Let me know if that is clear :)

jacobhq: Ok yeah that makes sense. Should it just be an api route that tries to test all of the rules?

crates/tuono_lib/src/env.rs
jacobhq: Do you mean some method on the env struct that wraps EnvManager, and same for writing files?

crates/tuono_lib/tests/env.rs
Valerioageno: Just api route that covers one single rule. Just to be sure that the whole flow is covered.

crates/tuono_lib/src/env.rs
Valerioageno: in the `MockEnv` struct that keeps track of the set files and system vars. This in order to cleanup just what was declarated

jacobhq: Ok, that sounds like a good idea

crates/tuono_lib/src/env.rs
```diff
@@ -0,0 +1,281 @@
...
+        let env_vars: HashMap<String, String> = env::vars().collect();
+
+        let mut manager = Self {
+            env_files,
+            system_env_names,
+            env_vars,
+        };
+
+        manager.reload_variables();
+        manager.load_into_env();
```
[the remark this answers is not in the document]
jacobhq: I agree, however in the future (after v1 maybe), I think hot reload could be greatly improved, and in that case we will need it to be a struct with methods.

crates/tuono/src/watch.rs
```diff
@@ -114,6 +122,15 @@ pub async fn watch() -> Result<()> {
...
                 }
+
+                if path
+                    .0
+                    .file_name()
+                    .map(|f| f.to_string_lossy().starts_with(".env"))
+                    .unwrap_or(false)
+                {
+                    should_reload_ssr_bundle = true
+                }
```
Valerioageno, line 133: Better export this condition to a function and merge it with the other `should_reload_rust_server` logic. This for keeping a single source of thruth

crates/tuono_lib/src/env.rs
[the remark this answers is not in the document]
jacobhq: I don't think it should be a function, since even if we are not using the struct idiomatically at the moment. For now, let's:

- Move `load_into_env` into `reload_variables` to avoid looping twice
- Rename `reload_variables` to `load_variables` for now (but stays easy to rename later if its scope changes)

WDYT?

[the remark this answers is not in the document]
Valerioageno: I don't understand why it shouldn't be a function.
Its purpose is to load the variables on the process and return `()`. Why creating a struct for it?

> we are not using the struct idiomatically at the moment

Check out the [YAGNI principle](https://martinfowler.com/bliki/Yagni.html).

I'd create a single function (that could be split into sub-function for readability reasons).

WDYT?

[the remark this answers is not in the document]
jacobhq: Ok yeah, that's a good point. I'll split it out into a function :)

crates/tuono/src/source_builder.rs
```diff
@@ -213,10 +213,11 @@ mod tests {
     #[test]
     fn should_set_the_correct_mode() {
         let source_builder = App::new();
-
         let dev_bundle = generate_axum_source(&source_builder, Mode::Dev);
+
         assert!(dev_bundle.contains("const MODE: Mode = Mode::Dev;"));
 
+        let source_builder = App::new();
```
Valerioageno, line 220: why shadowing the previous App?

crates/tuono_lib/src/env.rs
```diff
@@ -0,0 +1,222 @@
...
+        mock_env.setup_env_file(".env", "TEST_KEY=base_value");
+        mock_env.setup_env_file(".env.development", "TEST_KEY=development_value");
+        mock_env.setup_env_file(".env.development.local", "TEST_KEY=local_dev_value");
+
+        load_env_vars(Mode::Dev);
+
+        mock_env.capture_keys(&["TEST_KEY"]);
+
+        assert_eq!(env::var("TEST_KEY").unwrap(), "local_dev_value");
+    }
```
Valerioageno, line 153: Can we do also a test with `Mode::Prod`?

crates/tuono/src/watch.rs
```diff
@@ -74,6 +77,14 @@ fn build_react_ssr_src() -> Job {
     .0
 }
 
+fn ssr_reload_needed(file_name: Option<&OsStr>, file_path: Cow<str>) -> bool {
+    file_name
+        .map(|f| f.to_string_lossy().starts_with(".env"))
+        .unwrap_or(false)
+        || file_path.ends_with("sx")
+        || file_path.ends_with("mdx")
+}
```
Valerioageno, line 86: Can we have this function taking only `path.0` as a single parameter? It is weird to me having it taking two arguments that are basically the same

Valerioageno (approved):
Happy to approve this PR :)

jacobhq:
Thanks for all the reviews :)

PR #584, opened 2025-02-21
feat(env): setup env prefix and typescript for client-side

### LINKED ISSUE #570: Enhancement: Implement .env file support for configuration management

### LINKED ISSUE #587: Add `TUONO_PUBLIC_` prefix to vite `.env` variables

Refer to -> https://github.com/tuono-labs/tuono/issues/570

Refer to -> https://github.com/tuono-labs/tuono/issues/570

_No response_

### Additional context

_No response_

jacobhq:
### Checklist

### Related issue

Fixes #587 

### Overview

Vite env prefix (`TUONO_PUBLIC_`), and set up support for env clientside (no extra work apart from fixing TS issues that exist at the moment), and write tests (fixing also the typescript errors)

- [x] Fix TS errors


examples/tuono-tutorial/tsconfig.json
```diff
@@ -24,7 +24,10 @@
     "noFallthroughCasesInSwitch": true,
 
     // Completeness
-    "skipLibCheck": true
+    "skipLibCheck": true,
```
jacobhq, line 27: This is the only example with this line. From the [vite docs](https://vite.dev/guide/features#typescript-compiler-options:~:text=alwaysStrict-,skipLibCheck,TypeScript.%20You%20can%20learn%20more%20at%20vuejs/vue%2Dcli%235688.,-Client%20Types), I don't think we need it.

marcalexiei: `skipLibCheck` is set in every example:

- https://github.com/tuono-labs/tuono/blob/d6fc18e2f685f9369506e465c452bfae09b554e3/examples/tuono-app/tsconfig.json#L7
- https://github.com/tuono-labs/tuono/blob/d6fc18e2f685f9369506e465c452bfae09b554e3/examples/tuono-tutorial/tsconfig.json#L27
- https://github.com/tuono-labs/tuono/blob/d6fc18e2f685f9369506e465c452bfae09b554e3/examples/with-mdx/tsconfig.json#L7
- https://github.com/tuono-labs/tuono/blob/d6fc18e2f685f9369506e465c452bfae09b554e3/examples/with-mdx/tsconfig.json#L7

---

If you want to remove `skipLibCheck` you should also perform a `tsc` run in each example to ensure that there are no broken types from libraries. 

---

@Valerioageno at the moment there is any code check running inside `examples/*`.
How about adding the following check for each example:

- run `typecheck` script which run `tsc --noEmit`
- run a `tuono build`

in a new CI workflow to ensure that examples can be build without any issues?

jacobhq: > skipLibCheck is set in every example

Sorry didn't see that! It was late :sweat_smile:. In that case we should keep it in.

I agree with CI stuff - I think for a separate issue :)

marcalexiei: #590

packages/tuono/client.d.ts
marcalexiei, line 1: This file won't be included when publishing `tuono` package to npm:

https://github.com/tuono-labs/tuono/blob/d6fc18e2f685f9369506e465c452bfae09b554e3/packages/tuono/package.json#L54-L59

The file should be part of the package build.

1. create a folder `build-client` (just a proposal, I don't think this should be part of `build` folder) 
2. as content of the `index.ts` add `import 'vite/client'`
3. add the new entry to the build setup
4. correct `package.json` exports path
5. build
6. check that examples css imports do not report a typechecking error

marcalexiei (changes requested):

Valerioageno:
@jacobhq can you write few lines about why you needed to export a type from `tuono/client`? It doesn't seem straightforward

jacobhq:
> @jacobhq can you write few lines about why you needed to export a type from tuono/client? It doesn't seem straightforward

Sure thing!

Vite provides some stuff (including env vars) on an object called `import.meta`. I hadn't encountered this object before, but after a brief read of [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta), I found it provides the module with context about itself like a URL etc.

Without this change, typescript had no idea about what was inside `import.meta`, and so treated it as you trying to get something that didn't exist (ts 2339: Property `env` does not exist on type `ImportMeta`.). After a quick google, I came across [this stackoverflow question](https://stackoverflow.com/questions/78114219/property-env-does-not-exist-on-type-importmeta-ts2339), which suggested creating a `vite-env.d.ts` with the following content (it also suggested another approach, I used that later on):

```ts
/// <reference types="vite/client" />
```

This works in vite projects, because vite exists as a dependency, but in tuono projects, vite is a dependency of tuono, so this doesn't work. I fixed this by importing client in `packages/tuono/client.d.ts`, exporting that as `tuono/client`, and then referencing that in `tsconfig.json` of new tuono projects. I took this approach because I hate all the files in the root of a project ([look at how beautiful and simple it used to be in nextjs](https://github.com/vercel/title-site)), using `tsconfig.json` so I didn't have to make a new root file in each project.

Hope that was useful. I had a long session last night trying to fix this, so it is fresh in my mind :)

packages/tuono/client.d.ts
jacobhq: Should I remove `client.d.ts` after doing this? Also should I put it in `src/types.ts`?

marcalexiei: > Should I remove client.d.ts after doing this? 

yes 

> Also should I put it in src/types.ts?

I don't understand this point. What change do you want to do in `src/types.ts`?

jacobhq: > as content of the index.ts add import 'vite/client'

Was thinking of putting it in with the rest of the types. Although now i think, it needs to be separate.

marcalexiei: Exactly, it needs to be separated from the other, otherwise all types declared in types will be available in every file

marcalexiei:
TL;DR;

By adding `tuono/client`, which aliases `vite/client`, to `tsconfig#types` we include a file 
which contains declarations about variables and files handled by `vite`:

- https://vite.dev/guide/env-and-mode#built-in-constants
- module declaration for
  - images
  - stylesheets (css, scss, less, ...)
  - various file types related to assets (mp3, ...)

Check `vite/client` content for more info: https://github.com/vitejs/vite/blob/main/packages/vite/client.d.ts

---

The major benefit of this is that we avoid users creating this kind of file inside their apps.
E.g.: I have done this myself in [examples/tuono-tutorial/src/css-modules.d.ts](https://github.com/tuono-labs/tuono/blob/main/examples/tuono-tutorial/src/css-modules.d.ts)

---

Kudos to @jacobhq for resolving this using this approach!

packages/tuono/src/client-types/index.ts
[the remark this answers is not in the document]
Valerioageno: What about having `tuono/typescript`? Or `tuono/tsconfig`.

I think is pretty meaningful for the place in which it is used

Valerioageno:
Thanks @jacobhq for the explanation. Great work with that!

packages/tuono/src/client-types/index.ts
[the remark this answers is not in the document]
jacobhq: What if we just called it `vite-client-types` or something like that? It makes it very clear what it is.

jacobhq:
No worries :)

packages/tuono/src/client-types/index.ts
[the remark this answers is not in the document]
marcalexiei: Since both of you have proposed alternatives to `build-client`, 
feel free to choose whatever you think is best, 
as long as we have a `src` folder that matches the `exports` path, 
to maintain consistency with the current package structure.

However, I have some concerns about your proposals:

> tuono/typescript

It seems more like a fork of TypeScript rather than an export to be used inside `compilerOptions.types`.
Additionally, it appears that this should contain more type information, 
but right now, it only augments import.meta and adds many module declarations.

> tuono/tsconfig

This is misleading. 
It will be used inside `compilerOptions.types`, so it will resolve to a declaration file, not an actual tsconfig.

> vite-client-types

In the future, we might add custom declarations or make changes on the vite side, 
so I would avoid including vite in the name.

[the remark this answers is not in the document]
Valerioageno: All your points are fair! What about `compiler-types`?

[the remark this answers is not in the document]
marcalexiei: The term `compiler-types` sounds off to me, similar to how `tuono/typescript` feels. 
The name implies it should contain more type information, 
but right now, it primarily augments `import.meta` and adds module declarations.

It seems like it's simply a way of saying "put this inside `compilerOptions.types,`" 
but it’s more closely tied to the build environment than to any of the types provided by `tuono`.

If you want to go this route, `bundler` might be a better fit 
since these types are defining the bundler environment,
so `bundler-types` could make more sense.

However, I still lean toward `build-client`, 
as these types are more directly related to the build process 🫠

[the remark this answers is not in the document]
jacobhq: Shall we just call it `build-client` then :)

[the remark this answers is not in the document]
Valerioageno: Let's go with `build-client`

Valerioageno (approved):
LGTM

marcalexiei (approved):

PR #595, opened 2025-02-23
feat: update axum to v0.8.1

Valerioageno:
### Checklist

### Related issue

Fixes None

### Overview

Supersedes https://github.com/tuono-labs/tuono/pull/495

### DIFF
--- a/crates/tuono/src/route.rs
+++ b/crates/tuono/src/route.rs
@@ -46,6 +46,20 @@ impl AxumInfo {
         }
 
         if route.is_dynamic {
+            let dyn_re = Regex::new(r"\[(.*?)\]").expect("Failed to create dyn regex");
+            let catch_all_re =
+                Regex::new(r"\{\.\.\.(.*?)\}").expect("Failed to create catch all regex");
+
+            let dyn_result = dyn_re.replace_all(&axum_route, |caps: &regex::Captures| {
+                format!("{{{}}}", &caps[1])
+            });
+
+            let axum_route = catch_all_re
+                .replace_all(&dyn_result, |caps: &regex::Captures| {
+                    format!("{{*{}}}", &caps[1])
+                })
+                .to_string();
+
             return AxumInfo {
                 module_import: module
                     .as_str()
@@ -55,10 +69,7 @@ impl AxumInfo {
                     .replace('[', "dyn_")
                     .replace("...", "catch_all_")
                     .replace(']', ""),
-                axum_route: axum_route
-                    .replace("[...", "*")
-                    .replace('[', ":")
-                    .replace(']', ""),
+                axum_route,
             };
         }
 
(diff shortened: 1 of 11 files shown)


crates/tuono/src/route.rs
```diff
@@ -46,6 +46,20 @@ impl AxumInfo {
         }
 
         if route.is_dynamic {
+            let dyn_re = Regex::new(r"\[(.*?)\]").expect("Failed to create dyn regex");
```
jacobhq, line 49: Regex can be quite slow. It is better to use the previous rust code, since we are matching quite simple patterns.

This is probably negligible for 90% of apps, but it may make a difference for really big apps. WDYT?

crates/tuono/src/route.rs
```diff
@@ -46,6 +46,20 @@ impl AxumInfo {
...
+            let catch_all_re =
```
jacobhq, line 50: I do like the idea of adding some objects that represent each match, but IMHO these should be functions or closures instead of regex. WDYT

jacobhq (changes requested):
Looks good! 

Regex is bad for performance and can be hard to read. Since we are matching simple cases it is faster to keep using `.replace()`.

crates/tuono/src/route.rs
Valerioageno: Thanks for having raised this point. I had pretty meaningful readings on [regex crate](https://crates.io/crates/regex) documentation (it is quite a rabbit hole 😄 ).

Are we sure it is slower? Doing multiple string replacements means multiple lookups and multiple allocations so I'm not sure how much more performant it could be.
Also the regex complexity depends on the passed strings and I'm pretty confident that no strings will be longer than few tens (maybe hundreds) of characters.

For the same discussion  we had on the `.env` feature the `tuono` CLI is a software that is ok not being top performant. I would prioritize code readability.

Definitely I could build the regex lazily to avoid computation waste on the loop. WDYT?

crates/tuono/src/route.rs
Valerioageno: I haven't understood! Could you do an example?
Like a method of `AxumInfo`?

crates/tuono/src/route.rs
jacobhq: Yeah ok. Probably just comes down to readability here then, so maybe it's fine.

> Definitely I could build the regex lazily to avoid computation waste on the loop.

Yeah that would be good, although to be honest, if performance doesn't matter too much, you should only do it if it's simple :)

crates/tuono/src/route.rs
jacobhq: Oh wait, I think this would be too much abstraction. I meant for the replacement in the `axum_route` field. Definitely not needed 😅

crates/tuono/src/route.rs
Valerioageno: I'm resolving for now! The implementation would be quite straightforward but I'd have to load a separate crate for doing the lazy loading. Better keep it lean for now!

PR #607, opened 2025-02-26
feat: display error overlay

### LINKED ISSUE #592: Create `ErrorOverlay` component
## Overview

We need to create a custom overlay component to show React error on the browser rather than the terminal running the project.

<details><summary>In a next iteration, we should also be able to pass `rust` errors to it for enabling the full exceptions management from the browser</summary>
<p>

[Related task](https://github.com/orgs/tuono-labs/projects/2/views/2?filterQuery=-status%3ADone+&pane=issue&itemId=99040520)

</p>
</details> 

## Resources
- https://vite.dev/config/server-options.html#server-warmup
- https://vite.dev/config/server-options.html#server-hmr

Valerioageno:
### Checklist

### Related issue

Fixes #592 

#### Implementation detail
The dialog is a vite internal web component that gets automatically triggered by its WebSocket connection. 

In order to make this feature working I had to "hack" a little bit the current implementation
1. Since most of the compilation errors causes a failed SSR I created a `fallback_html` file that gets injected in `.tuono/index.html` that simply renders the project client side and open the connection to the vite server.
2. There is no a straighforward way to override the defaul vite error overlay. I followed the same implementation as astro (as suggested by one of the vite maintainers https://github.com/vitejs/vite/issues/19552)

#### Implementation split
1. Build `index.html` from `tuono` CLI
2. Build `error-overlay.ts` component in the vite build
3. Add e2e test for the overlay (this caused issues, will do it in a separate PR)

#### Screenshot
For this iteration the overlay is just a clone of the vite one. In the future we might need to refactor the UI in order to make it more `tuono` style like. No changes to the current PR should be required

Screenshot 2025-03-04 182324


crates/tuono/src/source_builder.rs
```diff
@@ -184,6 +202,30 @@ fn generate_axum_source(app: &App, mode: Mode) -> String {
...
+            let url = format!("http://{}:{}", config.server.host, config.server.port);
+            FALLBACK_HTML.replace("[BASE_URL]", url.as_str())
+        }
+    } else {
+        "".to_string()
+    }
+}
+
+pub fn generate_fallback_html(app: &App) -> io::Result<()> {
+    let base_path = std::env::current_dir().unwrap();
```
jacobhq, line 219: Let's not handle this with unwrap. Use expect until we standardise our errors more.

packages/tuono/src/build/error-overlay.ts
```diff
@@ -0,0 +1,299 @@
...
+`
+
+const HTMLElement: typeof globalThis.HTMLElement =
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  globalThis.HTMLElement ??
+  // eslint-disable-next-line @typescript-eslint/no-extraneous-class
+  class {}
+export class ErrorOverlay extends HTMLElement {
+  /**
+   * All the fields need to be implemented in the constructor otherwise
```
jacobhq, line 180: Why? Do you mean we can't add something like `private root: ShadowRoot;` here to prevent all the ts ignores?

packages/tuono/src/build/error-overlay.ts
jacobhq, line 1: Overall I think you are using too many ts/linter ignore comments here. Can we just add fields like vite did?

https://github.com/vitejs/vite/blob/2c51565ec044904a080ef5649034c37f02212c7b/packages/vite/src/client/overlay.ts#L210-L211

crates/tuono_lib/src/ssr.rs
jacobhq, line 1: Looks good. Can we give more detailed errors (it's annoying because they will probably never occur, but it's better than panicking IMHO)

examples/tuono-app/src/routes/index.tsx
jacobhq, line 1: Please undo the whitespace changes to examples.

jacobhq (changes requested):
Overall this looks pretty good. I don't understand the removal of the class fields on the ts side. Would you be able to tell me more about that?

examples/tuono-tutorial/src/components/PokemonLink.tsx
```diff
@@ -1,7 +1,6 @@
 // src/components/PokemonLink.tsx
 import type { JSX } from 'react'
 import { Link } from 'tuono'
-
```
marcalexiei, line 4: Revert this

packages/tuono/src/build/error-overlay.ts
marcalexiei: when specifying class attributes `esbuild` add some helper that cause runtime error when overlay is displayed:

image

I'm taking care of this.

packages/tuono/src/build/error-overlay.ts
marcalexiei: See reply on file comment

packages/tuono/src/build/error-overlay.ts
```diff
@@ -171,21 +173,17 @@ const overlayTemplate = `
 `
 
 const HTMLElement: typeof globalThis.HTMLElement =
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-  globalThis.HTMLElement ??
-  // eslint-disable-next-line @typescript-eslint/no-extraneous-class
-  class {}
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition, @typescript-eslint/no-extraneous-class
+  globalThis.HTMLElement ?? class {}
```
jacobhq, line 177: Sorry if this is a silly question, but can we just revert back to what vite used and then remove lint comment?

The only difference I can see is that our one works when `globalThis.HTMLElement` is `null`, however I think in nodejs, `globalThis.HTMLElement` [is just `undefined`](https://nodejs.org/api/globals.html), not explicitly set to `null`.

```ts
const { HTMLElement = class {} as typeof globalThis.HTMLElement } = globalThis
```

marcalexiei: No.
This code doesn't compile since we are using [isolatedDeclarations](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#isolated-declarations).

Try writing your proposed code and then run `pnpm run build`.

crates/tuono_lib/src/ssr.rs
Valerioageno: What do you mean with `more detailed errors`? Do you have any suggestion?

crates/tuono/src/source_builder.rs
Valerioageno: Nice catch! I updated it (I used the `error!` macro even though is not widely used here)

Valerioageno:
@marcalexiei @jacobhq Updated! Let me know if I can merge

marcalexiei (approved):

crates/tuono_lib/src/ssr.rs
jacobhq: I meant in `crates/tuono_lib/src/ssr.rs` there are a few unwraps. It would be nice to just handle those, and then that's all from me :)

Valerioageno: Updated (only for the dev mode)

jacobhq (approved):
All good from me. Nice work!

marcalexiei (approved):

PR #618, opened 2025-03-02
feat:  Implement .env file support

### LINKED ISSUE #570: Enhancement: Implement .env file support for configuration management

marcalexiei:
### Checklist

### Related issue

Fixes #570

### Overview

See #570


marcalexiei (commented):
@jacobhq / @Valerioageno all #570 sub tasks are completed .

Just noticed that in the old PR (#581) there was a "E2E test to check consistency between client and server" task.
I assume that it is blocked by #625, so I think we have 2 options:
1.  use the "base" fixture and write them now
2. create a separate task and perform it after #625 is resolved

Valerioageno:
I'd use the `base` fixture. In the end we just need to add the `.env` file and check that the value is correctly read.

Valerioageno:
The plan was in @jacobhq's POC. In my opinion we should not add any `e2e` test.
Our server implementation is heavily tested on `tuono_lib`. The client side is a vite feature so it does not need to be tested on our side.

I would avoid overtesting right now since we might need to do big changes and this would reduce our agility. WDYT?

jacobhq:
Would a vite change to env behaviour be breaking? If so, then I don't think we need e2e as much.

I do see it's value though (a vite change wouldn't cause any tests to fail, so it would probably get through code review), although I don't know how I would implement it since I haven't used playwright before.

crates/tuono/src/watch.rs
```diff
@@ -86,6 +98,13 @@ pub async fn watch() -> Result<()> {
 
     let build_ssr_bundle = build_react_ssr_src();
 
+    let env_files = fs::read_dir("./")
+        .unwrap()
```
jacobhq, line 102: Looks like I left in an unwrap here :)

```suggestion
        .expect("Error reading env files from current directory")
```

marcalexiei:
> In my opinion we should not add any e2e test.
> Our server implementation is heavily tested on tuono_lib. 
> The client side is a vite feature so it does not need to be tested on our side.

From my point of view:

- the two implementations are heavily intertwined.
- client-side implementation  `vite` implementation is completely not under tuono control.

so they should be tested to ensure their behave the same.

---

> I would avoid overtesting right now since we might need to do big changes and this would reduce our agility.

I understand your point, but I have a different perspective. 
If we avoid testing features because _we might need to make big changes_
we risk delivering new features that could break existing ones.

## Closing throughs regarding env feature

I understand that you are eager to release the env feature,
and `vite` env implementation is not like to not change soon,
so, as a compromise, I propose to create a new issue to test this feature 
(labelled as good first issue like  #625).

jacobhq:
> so, as a compromise, I propose to create a new issue to test this feature

Yes that is a great idea. This will definitely need e2e at some point (before v1) IMHO.

marcalexiei:
A new issue has been created: #634

Valerioageno:
Thanks for it! I agree on the compromise

Valerioageno (approved):
I QA the feature and it works as expected ✅

PR #629, opened 2025-03-08
refactor(packages/tuono-router): use named imports

### LINKED ISSUE #259: [Repo maintenance]: use named import when working with React
> Style: useState and useEffect are imported via named import while all context functions are prefixed with React..
I think we should be more consistent.
I usually prefer importing React and prefix all with React. (also properties which starts with React. It's redundant, but at least the style is consistent on all files).
There are types like ElementType or handlers (like MouseEventHandler, ...) that cannot be associated with React when used without React.

> I'd always use named imports.

_Originally posted by @marcalexiei and @Valerioageno in https://github.com/tuono-labs/tuono/pull/256#discussion_r1896550715_

marcalexiei:
### Checklist

### Related issue

Fixes #259

### Overview

- See linked issue
- Replaced some default export with named export (does not impact usage on user-side)
  - `<NotFound />` 
  - `<Link />`

### DIFF
--- a/packages/tuono-router/src/components/Link.spec.tsx
+++ b/packages/tuono-router/src/components/Link.spec.tsx
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest'
 import { render, fireEvent, screen } from '@testing-library/react'
 
-import Link from './Link'
+import { Link } from './Link'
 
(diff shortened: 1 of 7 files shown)


Valerioageno (approved):
LGTM! Thanks

PR #630, opened 2025-03-08
chore: add `@vitest/eslint-plugin`

### LINKED ISSUE #296: [Repo maintenance]: ESLint - add `@vitest/eslint-plugin`
Vitest has its dedicated `eslint` plugin which is basically a port of `eslint-plugin-jest`:

https://www.npmjs.com/package/@vitest/eslint-plugin

Since we are using `vitest` we should add this plugin to our eslint configuration.

marcalexiei:
### Checklist

### Related issue

Fixes #296

### Overview

Add `@vitest/eslint-plugin` to eslint configuration.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -5,9 +5,10 @@ import eslintPluginImport from 'eslint-plugin-import'
 import eslintPluginReact from 'eslint-plugin-react'
 // @ts-expect-error no types are available for this plugin
 import eslintPluginReactHooks from 'eslint-plugin-react-hooks'
+import eslintPluginVitest from '@vitest/eslint-plugin'
 
-const REACT_FILES_MATCH =
-  'packages/{tuono,tuono-router}/**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'
+const JS_EXTENSIONS_MATCH = 'js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx'
+const REACT_FILES_MATCH = `packages/{tuono,tuono-router}/**/*.{${JS_EXTENSIONS_MATCH}}`
 
 /** @type import('typescript-eslint').ConfigArray */
 const tuonoEslintConfig = tseslint.config(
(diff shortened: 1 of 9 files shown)


eslint.config.js
```diff
@@ -5,9 +5,10 @@ import eslintPluginImport from 'eslint-plugin-import'
 import eslintPluginReact from 'eslint-plugin-react'
 // @ts-expect-error no types are available for this plugin
 import eslintPluginReactHooks from 'eslint-plugin-react-hooks'
+import eslintPluginVitest from '@vitest/eslint-plugin'
 
-const REACT_FILES_MATCH =
-  'packages/{tuono,tuono-router}/**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'
+const JS_EXTENSIONS_MATCH = 'js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx'
```
Valerioageno, line 10:
```suggestion
const JS_EXTENSIONS_MATCH = ['js','mjs','cjs','jsx','mjsx','ts','tsx','mtsx'].join(',')
```

eslint.config.js
```diff
@@ -5,9 +5,10 @@ import eslintPluginImport from 'eslint-plugin-import'
...
+const REACT_FILES_MATCH = `packages/{tuono,tuono-router}/**/*.{${JS_EXTENSIONS_MATCH}}`
```
Valerioageno, line 11: What about making it enabled for all the packages? In case we need a new package, we risk forgetting about it

```suggestion
const REACT_FILES_MATCH = `packages/**/*.{${JS_EXTENSIONS_MATCH}}`
```

marcalexiei: I don’t understand why we should apply React rules to packages that aren't using React, like `tuono-fs-router-vite-plugin`. 
Running these rules on packages without `react` is a waste of resources and also issues a warning since the rules are unable to detect the react version to be used inside rules check.

Valerioageno: Yeah I agree. I don't have a better idea right now about preventing to miss updating this config. Let's go this way!

eslint.config.js
marcalexiei: I don't see any particular advantage of this, but I don't have a strong opinion so I will apply your version.

Issue #617, opened 2025-03-02
[Bug]: link leading to 404 cause a blank page to be displayed


marcalexiei:
### Description

When user clicks on a `<Link />` targeting  a not existing route a blank page is displayed.

> [!WARNING]
> The issue appears only when clicking on a Tuono `<Link />`,
> loading the page directly or accessing it via HTML `<a />` works as expected

### Expected behaviour

404 default page is displayed

### How to reproduce

- Clone https://github.com/marcalexiei/tuono-repro-tutorial-404
- `cargo install tuono@0.17.10`
- `pnpm i`
- `tuono dev`
- Open http://localhost:3000
- Click on a pokemon link

### System Info

```shell
System:
    OS: macOS 15.3.1
    CPU: (8) arm64 Apple M3
    Memory: 94.86 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.14.0/bin/npm
    pnpm: 9.15.5 - ~/.nvm/versions/node/v22.14.0/bin/pnpm
  Browsers:
    Chrome: 133.0.6943.55
    Safari: 18.3
  npmPackages:
    tuono: 0.17.10 => 0.17.10
```

### System info (Rust)

```shell
rustc 1.83.0 (90b35a623 2024-11-26)
cargo 1.83.0 (5ffbef321 2024-10-29)
tuono 0.17.10
```

### Additional context

https://github.com/tuono-labs/tuono-documentation/pull/43#discussion_r1976562818

PR #633, opened 2025-03-08
feat: add __layout wrapper to default 404 page

Valerioageno:
### Checklist

### Related issue

Fixes #0


packages/tuono-router/src/components/NotFound.tsx
```diff
@@ -15,10 +15,14 @@ export function NotFound(): JSX.Element {
     return <RouteMatch route={custom404Route} serverInitialData={{}} />
   }
 
+  const RootLayout = router.routesById['__root__']?.component
+
+  if (!RootLayout) return null
+
   return (
-    <>
+    <RootLayout data={{}} isLoading={false}>
```
marcalexiei, line 23: Since there is no data, why not passing `null` instead of `{}`?

```suggestion
    <RootLayout data={null} isLoading={false}>
```

marcalexiei (commented):
> might also fix https://github.com/tuono-labs/tuono/issues/617

Tested on `examples/tuono-tutorial` using local tuono build and the issue seems no longer present

Valerioageno:
All fair points. Applied!

PR #641, opened 2025-03-10
test(packages/tuono-router): improve test using mocks

marcalexiei:
### Checklist

### Related issue

Follow up of https://github.com/tuono-labs/tuono/pull/633#pullrequestreview-2669394172

### Overview

- uniform comonent style title ("Test ComponentName Component" → "< ComponentName />") 
- `packages/tuono-router/src/components/Link.spec.tsx`
  - add missing `cleanup` function 
- `packages/tuono-router/src/components/NotFound.spec.tsx`
  - use `vi.mocked` instead of mock + `vi.hoised`
  - mock components and hooks directly used by `NotFound.spec.tsx` 
  - replaced snapshots checks with mock checks to ensure component is working correctly
  - In order to have a consistent test system checks I created `NotFoundDefaultContent` instead of using snapshots or rtl checks
- `packages/tuono-router/src/components/RouteMatch.spec.tsx`
  - simplify `createRouteComponent` function
- `packages/tuono-router/src/hooks/useRoute.spec.ts`
  - use `vi.mocked` instead of mock + `vi.hoised`
  - remove redundant check on current route

### DIFF
--- a/packages/tuono-router/src/components/NotFound.spec.tsx
+++ b/packages/tuono-router/src/components/NotFound.spec.tsx
@@ -1,114 +1,88 @@
-import type { HTMLAttributes, JSX } from 'react'
+import type { ReactNode } from 'react'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 
-import { cleanup, render, screen } from '@testing-library/react'
+import { cleanup, render } from '@testing-library/react'
 
 import { Route } from '../route'
-import type { RouteComponent, RouteProps } from '../types'
+import type { RouteComponent } from '../types'
+
+import type { RouterInstanceType } from '../router'
 
 import { NotFound } from './NotFound'
-
-function createRouteComponent(
-  routeType: string,
-  RouteComponentFn: (props: RouteProps) => JSX.Element,
-): RouteComponent {
-  const RootComponent = RouteComponentFn as RouteComponent
-  RootComponent.preload = vi.fn()
-  RootComponent.displayName = routeType
-  return RootComponent
-}
-
-const root = new Route({
-  isRoot: true,
-  component: createRouteComponent('root', ({ children }) => (
-    <div data-testid="root">{children}</div>
-  )),
-})
-
-vi.mock('./Link', () => ({
-  Link: (props: HTMLAttributes<HTMLAnchorElement>): JSX.Element => (
-    <a {...props} />
-  ),
-}))
-
-vi.mock('../hooks/useServerPayloadData.ts', () => ({
-  useServerPayloadData: (): { data: unknown; isLoading: boolean } => {
-    return {
-      data: undefined,
-      isLoading: false,
-    }
-  },
-}))
-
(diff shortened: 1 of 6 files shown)


packages/tuono-router/src/components/NotFoundDefaultContent.tsx
```diff
@@ -0,0 +1,12 @@
+import type { JSX } from 'react'
+
+import { Link } from './Link'
+
+export function NotFoundDefaultContent(): JSX.Element {
+  return (
+    <>
+      <h1>Page Not Found</h1>
+      <Link href="/">Return to Homepage</Link>
```
marcalexiei, line 9: This is the only user side relevant change:

| Element | Previous | New |
|--------|--------|--------|
| **Title** | 404 Not found | Page Not Found |
| **Link text** | Return home | Return to Homepage |

> [!NOTE]
> Since is just a wording change I think that it was overkill to create a separate PR, 
> let me know if you prefer I perform this change separately.

packages/tuono-router/src/components/NotFound.spec.tsx
```diff
@@ -1,114 +1,86 @@
...
+      useRouterContextMock.mockReturnValue({
         router: {
           routesById: {
             __root__: root,
           },
         },
       })
+
       render(<NotFound />)
-      expect(screen.getByTestId('root')).toMatchInlineSnapshot(
```
Valerioageno, line 96: The goal was asserting that when no custom 404 page was provided the default was wrapped in the user root. Why are we removing this assertion?

packages/tuono-router/src/components/NotFoundDefaultContent.tsx
Valerioageno: That is fine to me. Less bureaucracy

Valerioageno: Looks overkill tho having a separate file for the content. If the reason is only easier testing mocking I would not do it.

packages/tuono-router/src/components/NotFound.spec.tsx
marcalexiei: I used mock to test this behaviour, if `RouteMatchMock` is called than it means that the root layout is present

marcalexiei: I added `rootRouteComponentMock` and relevant checks to restore the previous checks

packages/tuono-router/src/components/NotFoundDefaultContent.tsx
Valerioageno: What do you think about this?

marcalexiei: I prefer this approach because keeps the default render separated from the pure logic component.

Valerioageno (approved):
LGTM

PR #646, opened 2025-03-14
fix: prevent returning the previous route data on route transition

### LINKED ISSUE #626: [Bug]: Routes will initially render with props from previous route
### Description

When navigating from one route to another route using the `Link` component from Tuono, the following props are provided to the route component:

- `{ "data": { <<< data from previous route >>>, "isLoading": false }`
- `{ "isLoading": true }`
- `{ "data": { <<< data for current route >>>, "isLoading": false }`

This issue appears to affect all route transition types:
- Regular to regular and reverse
- Regular to catch single and reverse
- Regular to catch all and reverse

The issue appears in both development and release modes.

### Expected behaviour

It is expected that the data from the previous route is not provided to the new route, as it does not match the expected data or type for the props of the new route. The follow props are expected when navigating to a new route:

- `{ "isLoading": true }`
- `{ "data": { <<< data for current route >>>, "isLoading": false }`

### How to reproduce

Nix is required for easy reproduction, see: https://nixos.org/download/

1. Build and run the Tuono server:

```
git clone --branch issue_catch_all_nav https://github.com/pveierland/tuono.git
cd tuono
nix develop
pnpm install && pnpm build
cd examples/tuono-app && nix run ../../.#tuono -- build && cargo run --release
```

2. Open `curl http://localhost:3000` in a browser.
3. Click `Catch None: Blastoise` link.
4. The bottom of the screen will show the props provided to the new route:
```
{"data":{"subtitle":"The react / rust fullstack framework"},"isLoading":false}
{"data":{"subtitle":"The react / rust fullstack framework"},"isLoading":true}
{"data":{"name":"blastoise","id":9,"weight":855,"height":16},"isLoading":false}
```

This shows that the props from the previous route is provided to the new route, which is not expected behavior.

### Screenshots

_No response_

### System Info

```shell
System:
    OS: Linux 6.6 NixOS 24.05 (Uakari) 24.05 (Uakari)
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-1370P
    Memory: 39.47 GB / 62.45 GB
    Container: Yes
    Shell: 5.2.37 - /nix/store/p6k7xp1lsfmbdd731mlglrdj2d66mr82-bash-5.2p37/bin/bash
  Binaries:
    Node: 20.18.1 - /nix/store/wfxq6w9bkp5dcfr8yb6789b0w7128gnb-nodejs-20.18.1/bin/node
    Yarn: 1.22.22 - /nix/store/jcgdksj946l5l42c2y9ks2l4g6n74h3f-yarn-1.22.22/bin/yarn
    npm: 10.8.2 - /nix/store/wfxq6w9bkp5dcfr8yb6789b0w7128gnb-nodejs-20.18.1/bin/npm
    pnpm: 9.14.4 - /nix/store/rzxqj4y899lcr3c2rx92pf457dk6086w-pnpm-9.14.4/bin/pnpm
  npmPackages:
    tuono: link:../../packages/tuono => 0.17.10
```

### System info (Rust)

```shell
rustc 1.83.0 (90b35a623 2024-11-26)
cargo 1.83.0 (5ffbef321 2024-10-29)
tuono 0.17.10
```

### Additional context

_No response_

Valerioageno:
### Checklist

### Related issue

Fixes #626

### DIFF
--- a/packages/tuono-router/src/components/RouteMatch.spec.tsx
+++ b/packages/tuono-router/src/components/RouteMatch.spec.tsx
@@ -46,16 +46,14 @@ const route = new Route({
 vi.mock('../hooks/useServerPayloadData', () => ({
   useServerPayloadData: vi.fn(),
 }))
-vi.mocked(useServerPayloadData).mockReturnValue({
-  data: undefined,
-  isLoading: false,
-})
+
+const useServerPayloadDataMock = vi.mocked(useServerPayloadData)
 
 describe('<RouteMatch />', () => {
   afterEach(cleanup)
 
   it('should correctly render nested routes', () => {
-    vi.mocked(useServerPayloadData).mockReturnValue({
+    useServerPayloadDataMock.mockReturnValue({
       data: { some: 'data' },
       isLoading: false,
     })
@@ -84,7 +82,7 @@ describe('<RouteMatch />', () => {
   })
 
   it('should return null data when while loading', () => {
-    vi.mocked(useServerPayloadData).mockReturnValue({
+    useServerPayloadDataMock.mockReturnValue({
       data: { some: 'data' },
       isLoading: true,
     })
--- a/packages/tuono-router/src/components/RouteMatch.spec.tsx
+++ b/packages/tuono-router/src/components/RouteMatch.spec.tsx
@@ -7,15 +7,21 @@ import { useServerPayloadData } from '../hooks/useServerPayloadData'
 
 import { RouteMatch } from './RouteMatch'
 
-function createRouteComponent(
-  routeType: string,
-  includeChildren: boolean,
-): RouteComponent {
+function createRouteComponent(routeType: string): RouteComponent {
   const RootComponent = (({ children }: RouteProps) => (
     <div data-testid={routeType}>
       {`${routeType} route`}
-      {includeChildren ? children : null}
+      {children}
     </div>
+  )) as RouteComponent
+  RootComponent.preload = vi.fn()
+  RootComponent.displayName = routeType
+  return RootComponent
+}
+
+function createLeafRouteComponent(routeType: string): RouteComponent {
+  const RootComponent = (({ data }: RouteProps) => (
+    <div data-testid={routeType}>{data ? JSON.stringify(data) : null}</div>
   )) as RouteComponent
   RootComponent.preload = vi.fn()
   RootComponent.displayName = routeType
@@ -24,16 +30,16 @@ function createRouteComponent(
 
 const root = new Route({
   isRoot: true,
-  component: createRouteComponent('root', true),
+  component: createRouteComponent('root'),
 })
 
 const parent = new Route({
-  component: createRouteComponent('parent', true),
+  component: createRouteComponent('parent'),
   getParentRoute: (): Route => root,
 })
 
 const route = new Route({
-  component: createRouteComponent('current', false),
+  component: createLeafRouteComponent('current'),
   getParentRoute: (): Route => parent,
 })
 
@@ -49,6 +55,11 @@ describe('<RouteMatch />', () => {
   afterEach(cleanup)
 
   it('should correctly render nested routes', () => {
+    vi.mocked(useServerPayloadData).mockReturnValue({
+      data: { some: 'data' },
+      isLoading: false,
+    })
+
     render(<RouteMatch route={route} serverInitialData={{}} />)
 
     expect(screen.getByTestId('root')).toMatchInlineSnapshot(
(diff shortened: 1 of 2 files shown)


packages/tuono-router/src/components/RouteMatch.tsx
```diff
@@ -24,10 +24,16 @@ export const RouteMatch = ({
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const routes = useMemo(() => loadParentComponents(route), [route.id])
 
+  const routeData = isLoading ? null : data
```
marcalexiei, line 27: This change impacts `TuonoProps` typings:

https://github.com/tuono-labs/tuono/blob/bfd0041c6a1a14a427543f32ca80c632783004b4/packages/tuono/src/types.ts#L50-L53

The interface should become a union type to provide to typescript the correct anatomy of the runtime value, which allows a better static code analysis:

```ts
export type TuonoRouteProps<T> = { 
   data: null
   isLoading: true
 } | { 
   data: T
   isLoading: false
 } 
```

This change could be done together with:
- https://github.com/tuono-labs/tuono/issues/450

I can take care of this when I have time

packages/tuono-router/src/components/RouteMatch.spec.tsx
```diff
@@ -64,11 +75,38 @@ describe('<RouteMatch />', () => {
...
       </div>
     `,
     )
   })
+
+  it('should return null data when while loading', () => {
+    vi.mocked(useServerPayloadData).mockReturnValue({
+      data: { some: 'data' },
+      isLoading: true,
+    })
```
marcalexiei, line 90: the test title says: "should return null data when while loading"
however the mock is returning `data` set to `{ some: 'data' }` which isn't the correct value
```suggestion
    vi.mocked(useServerPayloadData).mockReturnValue({
      data: null,
      isLoading: true,
    })
```

jacobhq (commented):
Moved in https://github.com/tuono-labs/tuono/pull/646/files#r1996042548

packages/tuono-router/src/components/RouteMatch.spec.tsx
```diff
@@ -49,6 +55,11 @@ describe('<RouteMatch />', () => {
   afterEach(cleanup)
 
   it('should correctly render nested routes', () => {
+    vi.mocked(useServerPayloadData).mockReturnValue({
```
marcalexiei, line 58: Since each test now has its own `mockReturnValue`, 
the `vi.mocked(useServerPayloadData)` on line 49:

```ts
vi.mocked(useServerPayloadData).mockReturnValue({
  data: undefined,
  isLoading: false,
})
```

no longer has any effect.

You should replace it with a `useServerPayloadDataMock` variable:

```ts
const useServerPayloadDataMock = vi.mocked(useServerPayloadData)
```

and use it in each test instead of calling `vi.mocked` every time.

marcalexiei (changes requested):

marcalexiei:
@jacobhq if the comment is regarding a specific piece of code (like a function in this scenario ) you should create a thread instead of commenting directly on the PR.

https://tuono.dev/documentation/contributing/pull-requests#addressing-review-feedback

> Please [post your comments](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request) as line comments when possible so they can be threaded.

packages/tuono-router/src/components/RouteMatch.spec.tsx
```diff
@@ -7,33 +7,39 @@ import { useServerPayloadData } from '../hooks/useServerPayloadData'
...
-      {includeChildren ? children : null}
+      {children}
     </div>
   )) as RouteComponent
   RootComponent.preload = vi.fn()
   RootComponent.displayName = routeType
   return RootComponent
 }
 
+function createLeafRouteComponent(routeType: string): RouteComponent {
```
marcalexiei, line 22: From @jacobhq review comment https://github.com/tuono-labs/tuono/pull/646#pullrequestreview-2686349679

> Looks good, although I don't understand why we need the createLeafRouteComponent function, would you mind explaining?

packages/tuono-router/src/components/RouteMatch.spec.tsx
Valerioageno: The goal of the test is to return `null` even when `data` exists (due to a routing transition)

packages/tuono-router/src/components/RouteMatch.tsx
Valerioageno: Yeah, that would be great! I can create the union type with this PR and then you address the name separately. WDYT?

packages/tuono-router/src/components/RouteMatch.spec.tsx
Valerioageno: Agree

packages/tuono-router/src/components/RouteMatch.spec.tsx
Valerioageno: I think it makes sense having them separate, since passing data creates a weird HTML output that is weird dealing with.

packages/tuono-router/src/components/RouteMatch.tsx
marcalexiei: This is a breaking change since it will change the anatomy of `TuonoProps` completely.
Because of this I think this should be addressed in a different PR and have a different changeling entry.

Valerioageno: So you think we can merge this and then you address the other one?

marcalexiei: yes

marcalexiei: https://github.com/tuono-labs/tuono/issues/450 has been updated with new type refinements

marcalexiei (approved):

PR #649, opened 2025-03-18
Suggested a new example.

buraksenyurt:
Hello.

While experimenting with **Tuono**, I noticed that there were not enough examples in the examples folder. I thought an example that deals with page routing would be useful. I used [**Chuck Norris** jokes api](https://api.chucknorris.io/) to add fun to the example. The main page lists categories retrieved with **HTTP get**. When we click on any category, a random joke from that category is shown on the routed page _(again via simple HTTP Get request)_

Here is the sample screenshot from.

image

Thanks.


Valerioageno (commented):
Hey, nice to meet you. Thanks for the PR and for playing with tuono.

Don't you think that `tuono-tutorial` example already covers the same routing mechanism?

buraksenyurt:
Nice to meet you too. Yes, they have the same logic. Thanks for reviewing.

Issue #635, opened 2025-03-09
[Feature request]:  Create a `form_data()` method for the `Request` struct


Valerioageno:
### Is your feature request related to a problem? Please describe.

[Discord thread](https://discord.com/channels/1307411506735874208/1307692152037572678/1348046044625371248)

There is no clear way to implement form submissions using the native HTTP `form_data`.

### Describe the solution you'd like

I'd add a `request.form_data<FormData>()` method to the `Request` (similarly to the [body method](https://tuono.dev/documentation/server-utilities/request#body))

kulapoo:
Would like to work on this.

Valerioageno:
Thanks for the interest! Sure go ahead! ⚡

lifeiyu071:
> [@lifeiyu071](https://github.com/lifeiyu071) let's not mix the PR scopes. Will take a look about that separately我们不要混淆 PR 范围。我们将单独讨论这一点

okay

PR #651, opened 2025-03-20
feat: add form_data method into request

kulapoo:
### Checklist

- [ x ] I have read [Contributing > Pull requests](https://tuono.dev/documentation/contributing/pull-requests)

### Related issue

Fixes  [635](https://github.com/tuono-labs/tuono/issues/635)

### Overview

This update lets users easily extract form data from incoming requests

demo

https://github.com/user-attachments/assets/ad1249e5-47a4-4c2a-8d21-2a6e2ec0fa3f


kulapoo:
I'll take care of those failing.. Also I might still have to update the docs. Will get back to this later once Im available.

kulapoo:
let me know what else needed here to change, although I have some suggestions related to this in the future but I might create separate feature `issue` for it.

crates/tuono_lib/src/request.rs
```diff
@@ -37,6 +38,7 @@ pub struct Request {
     pub uri: Uri,
     pub headers: HeaderMap,
     pub params: HashMap<String, String>,
+    pub body: Option<Vec<u8>>,
```
Valerioageno, line 41: Why having `body` public?

crates/tuono_lib/src/request.rs
```diff
@@ -68,6 +80,42 @@ impl Request {
...
+        let body_str = match self.body_as_string() {
+            Ok(s) => s,
+            Err(e) => return Err(FormError::ParseError(e)),
+        };
+
+        let form_data = match serde_json::to_value(
+            form_urlencoded::parse(body_str.as_bytes())
+                .into_owned()
+                .collect::<HashMap<String, String>>(),
+        ) {
```
Valerioageno, line 107: I think this is inefficient. We translate first the the bytes into a string (cloning the boty) and then again into bytes.

Couldn't we just read the bytes?

crates/tuono_lib/src/request.rs
```diff
@@ -68,6 +80,42 @@ impl Request {
...
+            Ok(value) => value,
+            Err(e) => return Err(FormError::ParseError(e.to_string())),
+        };
+
+        let result: T = match serde_json::from_value(form_data) {
+            Ok(value) => value,
+            Err(e) => return Err(FormError::ParseError(e.to_string())),
+        };
+
+        Ok(result)
```
Valerioageno, line 117:
```suggestion
        return match serde_json::from_value::<T>(form_data) {
            Ok(value) => Ok(value),
            Err(e) => Err(FormError::ParseError(e.to_string())),
        };
```

crates/tuono_lib/Cargo.toml
```diff
@@ -31,6 +31,7 @@ regex = "1.10.5"
 either = "1.13.0"
 tower-http = {version = "0.6.0", features = ["fs"]}
 colored = "2.1.0"
+form_urlencoded = "1.2.1"
```
Valerioageno, line 34: What about using a crate compatible with `serde`?

https://docs.rs/serde_urlencoded/0.7.1/serde_urlencoded/

crates/tuono_lib/src/request.rs
```diff
@@ -45,6 +47,16 @@ impl Request {
             uri,
             headers,
             params,
+            body: None,
```
Valerioageno, line 50: I'd define its value using a `new(...)` parameter in the API macro to be more declarative where it is called

crates/tuono_lib/src/request.rs
```diff
@@ -82,6 +130,39 @@ impl From<serde_json::Error> for BodyParseError {
...
+            FormError::ServerError(_) => axum::http::StatusCode::INTERNAL_SERVER_ERROR,
+        };
+
+        let body = axum::Json(serde_json::json!({
+            "error": self.to_string(),
+        }));
+
+        (status, body).into_response()
+    }
+}
```
Valerioageno, line 164: I think this is a great initiative but I'd scope it with a separate PR.

Also we should use `axum_core` for implementing these traits since it has a more stable API 

https://docs.rs/axum-core/0.5.0/axum_core/response/trait.IntoResponse.html

Valerioageno (changes requested):
Thanks for this PR! Overall looks good but I left few comments. I'm available for discussing them.

I'd ask you also to create a test case in the integration `tests/`.

I'm all ears also for the suggestions you are talking above :)

Valerioageno:
PS I'm assigning this PR to you just for better visibility

Valerioageno:
Hey @kulapoo, please merge your PR with the `main` branch. I just update the rust edition of the whole repository

crates/tuono_lib/src/request.rs
kulapoo: does this mean

it will be like below with for parameters?

```
Request::new(
            Uri::from_static("http://localhost:3000"),
            HeaderMap::new(),
            HashMap::new(),
            None
        )
```

crates/tuono_lib/src/request.rs
kulapoo: alright

kulapoo:
> Hey @kulapoo, please merge your PR with the `main` branch. I just update the rust edition of the whole repository.
> 
> Also I fixed the `post()` method. There might be some conflicts but I reused part of your PR so you shouldn't find many conflicts. Sorry about that

ahh alribght,, I think I could just close this then

crates/tuono_lib/src/request.rs
```diff
@@ -70,12 +71,42 @@ impl Request {
             "Failed to read body",
         )))
     }
+
+    pub fn form_data<T>(&self) -> Result<T, (StatusCode, String)>
```
Valerioageno, line 75: Can you use the same pattern as the `body` method? I mean using `Deserialize` and return `BodyParseError`?
I think it is better leaving to the users the choice of what status code return

kulapoo: done.

Valerioageno:
Thanks, @kulapoo, for your contribution

PR #591, opened 2025-02-22
refactor(examples): uniform `tsconfig.json` files

marcalexiei:
### Checklist

- [x] I have read [Contributing > Pull requests](https://tuono.dev/documentation/contributing/pull-requests)

### Related issue

Followup of https://github.com/tuono-labs/tuono/pull/584#discussion_r1966472221

### Overview

All `tsconfig` examples files follow the structure provided in the [typescript tsconfig reference](https://www.typescriptlang.org/tsconfig/#references).

---

<details><summary>Resolved</summary>
<p>

**Additional consideration:**
We might consider to create a package with the default tuono typescript config.
Something like `@tuono/ts-config` or `tuono-ts-config`

If you agree, let me know and I'll create a new task so we can discuss this further in the future.

</p>
</details>

PR #703, opened 2025-04-04
feat: add automatic types generation

### LINKED ISSUE #463: [Feature request]: generate typescript types from react structs

You have to define the types in the frontend (react) and on the server (rust).

You could generate the frontend typings with a crate like ts-rs so you don't have to type them in the frontend. Instead just import them from the generated type file. 

_No response_

### Additional context

_No response_

Valerioageno:
### Checklist

### Related issue

Fixes #463

### Overview

This feature adds the support for automatically types generation. By just `#[derive(...)]` the `Type` trait, the `struct` or `enum` will be automatically translated to a typescript type and it will be available from the `"tuono/types"` module

#### How to QA locally
For the best QA experience, run the command `RUST_LOG=tuono=trace t dev`

#### What features are exposed?
Checkout the new documentation page for all the details https://github.com/tuono-labs/tuono-documentation/pull/89

#### Does this feature cover all the possible user usages?

No. This feature covers most of the user needs but doesn't implement complex cases. Since it is very hard to know beforehand what are all the users' needs (and implementing variants is quite straightforward) I consider this PR ready enough for starting using it.


examples/with-mdx/tsconfig.json
```diff
@@ -26,7 +26,10 @@
     "noEmit": true,
 
     // Completeness
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "paths": {
+      "tuono/types": ["./.tuono/types.ts"]
```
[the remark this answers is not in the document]
Valerioageno: I see. This would is just great!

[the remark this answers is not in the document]
Valerioageno: Done. Do you know if there is a way to expose a `tsconfig.json` from our package so the final users don't have to deal with it?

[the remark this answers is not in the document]
marcalexiei: You can

1. Publish a standalone package with the tsconfig (`@tuono/tsconfig` or `tuono-ts-config` to be consistent with current packages)
2. Expose the config from tuono (through an additional exports `tuono/ts-config`)

However `include` option is overridden if user specifies a value on his own:

https://www.typescriptlang.org/tsconfig/#extends 

> It’s worth noting that [files](https://www.typescriptlang.org/tsconfig/#files), [include](https://www.typescriptlang.org/tsconfig/#include), and [exclude](https://www.typescriptlang.org/tsconfig/#exclude) from the inheriting config file overwrite those from the base config file, and that circularity between configuration files is not allowed.

---

Having a sharable `tsconfig` could still be useful, as users won't need to define all the options explicitly."
(Past reference: https://github.com/tuono-labs/tuono/pull/591)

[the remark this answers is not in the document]
Valerioageno: I like option two. If you agree I'd open an issue for it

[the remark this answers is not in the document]
marcalexiei: - https://github.com/tuono-labs/tuono/issues/711

crates/tuono/src/commands/dev.rs
```diff
@@ -1,8 +1,10 @@
 use std::fs;
 use std::path::Path;
+use std::path::PathBuf;
```
jacobhq, line 3:
```suggestion
use std::path::{Path, PathBuf};
```

crates/tuono/src/commands/dev.rs
```diff
@@ -84,39 +91,58 @@ pub async fn watch(source_builder: SourceBuilder) -> Result<()> {
         let mut should_reload_ssr_bundle = false;
         let mut should_reload_rust_server = false;
         let mut should_refresh_axum_source = false;
+        let mut paths_to_refresh_types: Vec<PathBuf> = vec![];
+        let mut removed_files_from_types: Vec<PathBuf> = vec![];
```
jacobhq, line 95: Better to use `Vec::new()` than the macro when making an empty vector.

```suggestion
        let mut paths_to_refresh_types: Vec<PathBuf> = Vec::new();
        let mut removed_files_from_types: Vec<PathBuf> = Vec::new();
```

e2e/fixtures/base/src/routes/index.tsx
```diff
@@ -1,15 +1,12 @@
 import type { JSX } from 'react'
 import type { TuonoRouteProps } from 'tuono'
 import { Link } from 'tuono'
-
-interface IndexProps {
-  subtitle: string
-}
+import type { MyResponse } from 'tuono/types'
```
jacobhq, line 4: Maybe we should have tuono/generatedTypes or something like that in case we want to use `tuono/types` for our own types. WDYT?

crates/tuono/src/commands/dev.rs
```diff
@@ -58,7 +60,12 @@ pub async fn watch(source_builder: SourceBuilder) -> Result<()> {
 
     let env_files = detect_existing_env_files();
 
-    start_all_processes(process_manager.clone()).await;
+    unsafe {
+        // It is safe to call this function because here
+        // only one thread is running and the lock is not
+        // needed by any other thread.
+        start_all_processes(process_manager.clone()).await;
```
jacobhq, line 67: I've personally never used unsafe rust, but this looks fine to me.

crates/tuono/src/typescript/types_jar.rs
```diff
@@ -0,0 +1,95 @@
+use crate::typescript::FileTypes;
+use glob::glob;
+use std::fs::read_to_string;
+use std::path::{Path, PathBuf};
+use tracing::error;
+
+const TUONO_MACRO_TRAIT_NAME: &str = "Type";
```
jacobhq, line 7: Let's share this value in some kind of constants file, since it's duplicated in `utils.rs`

crates/tuono/src/typescript/types_jar.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[derive(Debug, Clone, Default)]
+pub struct TypesJar {
+    types: Vec<FileTypes>,
+}
+
+impl TypesJar {
+    pub fn new() -> Self {
+        Self { types: vec![] }
```
jacobhq, line 16:
```suggestion
        Self { types: Vec::new() }
```

crates/tuono/src/typescript/types_jar.rs
```diff
@@ -0,0 +1,95 @@
+use crate::typescript::FileTypes;
+use glob::glob;
+use std::fs::read_to_string;
+use std::path::{Path, PathBuf};
+use tracing::error;
+
+const TUONO_MACRO_TRAIT_NAME: &str = "Type";
+
+#[derive(Debug, Clone, Default)]
```
jacobhq, line 9: Do we need `Debug` for this?

```suggestion
#[derive(Clone, Default)]
```

crates/tuono/src/typescript/types_jar.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Removes all types from the jar that are
+    /// present in the provided `file_path`.
+    /// This function is triggered when a file is deleted
+    pub fn remove_file(&mut self, file_path: PathBuf) {
+        self.types.retain(|ttype| ttype.file_path != file_path);
+    }
+
+    pub fn refresh_file(&mut self, path: PathBuf) {
+        if let Ok(file_str) = read_to_string(&path) {
+            if file_str.contains(TUONO_MACRO_TRAIT_NAME) {
```
jacobhq, line 30: Should we do an early return if `!file_str.contains(TUONO_MACRO_TRAIT_NAME)` for readability?

crates/tuono/src/typescript/file_types.rs
```diff
@@ -0,0 +1,89 @@
+use super::parser;
+use super::utils::has_derive_type;
+use std::error::Error;
+use std::path::PathBuf;
+use tracing::trace;
+
+/// Represents all the valid typescript types found in a file.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
jacobhq, line 8: We may not need `Debug` in production
```suggestion
#[derive(Clone, PartialEq, Eq)]
```

jacobhq (commented):
Very impressive. Just a few minor tweaks, and questions on nesting...

crates/tuono/src/commands/dev.rs
Valerioageno: Can you elaborate this? Aren't they the same thing ([source](https://doc.rust-lang.org/src/alloc/macros.rs.html#42))?

e2e/fixtures/base/src/routes/index.tsx
Valerioageno: The rationale was:
- types does not increment the bundle size so there is no real need of creating a separate subpackage for them
- `tuono/types` is the more handy name I found to keep the user experience nice. A longer name sounds worse in my opinion. 
WDYT?

crates/tuono/src/commands/dev.rs
Valerioageno: The function does not need to be unsafe but I preferred to mark it as `unsafe` in order to be 10x careful if we need to refactor it (and forcing the dev to read the comment)

crates/tuono/src/typescript/types_jar.rs
Valerioageno: Agree

crates/tuono/src/commands/dev.rs
jacobhq: I think I just read it when I was learning rust - it doesn't really make a difference, but when there's no need to use a macro it's slightly better to use the function. This [reddit comment](https://www.reddit.com/r/rust/comments/fs1g5h/comment/flz0im5/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) sums it up quite well, but from the [rest of the discussion](https://www.reddit.com/r/rust/comments/fs1g5h/vec_or_vecnew/), it seems to be just preference. It's up to you :)

crates/tuono/src/commands/dev.rs
jacobhq: Oh ok, that does make a lot of sense!

crates/tuono/src/commands/dev.rs
jacobhq: No I don't think there is one, we would have to write one. I don't know if it's worth it, but i can investigate more if you like.

Valerioageno: I think investigating how we could extend our clippy config could worth it. But indeed it does not make sense for now at our size!

Valerioageno:
@marcalexiei and @jacobhq, I consider this feature finished. It will definitely need to be extended in the future to support all the possible cases, but as starting point should cover most of the user usages.

Let me know what you think about it, and please try it on your local env to see if you find any "awkwardness" that I might have miss. Thanks!

PR #713, opened 2025-04-05
fix: prevent multiple server address print

### LINKED ISSUE #712: [Bug]: Remove multiple title displayed with `tuono dev`
### Description

Image

### Expected behaviour

We should only display it the first time we load the server

### How to reproduce

just run `tuono dev` and update any rust file

### Screenshots

_No response_

### System Info

```shell
Not needed
```

### System info (Rust)

```shell
Not needed
```

### Additional context

_No response_

Valerioageno:
### Checklist

### Related issue

Fixes #712

### Proposed solution

Screenshot 2025-04-05 145140

### DIFF
--- a/crates/tuono_internal/src/tuono_println.rs
+++ b/crates/tuono_internal/src/tuono_println.rs
@@ -1,0 +1,11 @@
+#[macro_export]
+/// Log a message in the terminal using the custom tuono formatter.
+/// The messages printed with this macro should inform or
+/// guide the user.
+///
+/// The debug/error messages should be printed using the `tracing` crate
+macro_rules! tuono_println {
+    ($($arg:tt)*) => {{
+        println!("  {}", format!($($arg)*));
+    }};
+}
--- a/crates/tuono/Cargo.toml
+++ b/crates/tuono/Cargo.toml
@@ -21,6 +21,8 @@ clap = { version = "4.5.4", features = ["derive", "cargo"] }
 tracing = "0.1.41"
 tracing-subscriber = {version = "0.3.19", features = ["env-filter"]}
 miette = "7.2.0"
+
+colored = "2.1.0"
 watchexec = "5.0.0"
 watchexec-signals = "4.0.0"
 watchexec-events = "4.0.0"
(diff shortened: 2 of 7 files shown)


crates/tuono_internal/src/tuono_println.rs
marcalexiei, line 1: From my perspective the message should include more information. E.g.:

- Adding the time of the log at the beginning of the message
  ```text
  [20:23] Reloading...
  [20:23] Ready
  ``` 

- Add the area where the event happens. E.g.:
  - server
  - frontend client
  
  ```text
  [20:23] [server]  Reloading...
  [20:23] [server]  Ready
  ```

marcalexiei, line 1: An additional improvement is to clear the terminal screen and print only the most recent relevant messages,
In this case, it would show only the most recent "Reloading" and "Ready" messages.

This behaviour can be observed when running the vite dev command.

Valerioageno: mmm. Those are fair points. For the record the logs are:
- only server related and depend only on changing rust files
- In theory Reloading/Ready happens very close to each other. I'm gonna do a test adding the timing but not sure about it

Valerioageno: I definitely agree. The only issue is that right now these two log messages are managed by two different processes. Doing a screen clear is very difficult and error prone with the current setup.
I'm available to think instead on a solution like the `vite` development process where the whole terminal screen is managed by the process instead of just loggin lines

Valerioageno: I tried adding the formatted timing but in rust there is not std library for that. I have to import a new lib and I'm skeptical because we should add it also to the `tuono_lib` which is the user server.

For this reason I think that having just the two lines is enough (considering also that we could move to a more integrates solution in the future as described in the comment below completely managed by the CLI process).

marcalexiei: https://github.com/orgs/tuono-labs/projects/2?pane=issue&itemId=105574253

marcalexiei (approved):

PR #730, opened 2025-04-09
fix: old data kept on new route during transition

### LINKED ISSUE #626: [Bug]: Routes will initially render with props from previous route

Valerioageno:
### Checklist

### Related issue

Fixes #626

### Overview

The issue is that `isLoading` actually waits only for the server side data rather than the full transition.
To solve the problem, I created a `isTransitioning` state in the `RouterContext` that keep track of the full transition.

### DIFF
--- a/packages/tuono-router/src/components/RouteMatch.spec.tsx
+++ b/packages/tuono-router/src/components/RouteMatch.spec.tsx
@@ -59,14 +59,10 @@ describe('<RouteMatch />', () => {
   afterEach(cleanup)
 
   it('should correctly render nested routes', () => {
-    useServerPayloadDataMock.mockReturnValue({
-      data: { some: 'data' },
-    })
+    useServerPayloadDataMock.mockReturnValue({ data: { some: 'data' } })
 
-    // @ts-expect-error no need to define the full context here
-    useRouterContextMock.mockReturnValue({
-      isTransitioning: false,
-    })
+    // @ts-expect-error only isTransitioning is used by RouteMatch
+    useRouterContextMock.mockReturnValue({ isTransitioning: false })
 
     render(<RouteMatch route={route} serverInitialData={{}} />)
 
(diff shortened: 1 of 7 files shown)


packages/tuono-router/src/components/RouteMatch.spec.tsx
marcalexiei, line 1: There should be a dedicated test for testing `isTransitioning` changes.

marcalexiei (commented):
I tested my changes locally so I'm going to push them. No need to rewrite them.

Issue #753, opened 2025-04-21
[Feature request]: Add the ability to fetch client connection information


aig787:
### Is your feature request related to a problem? Please describe.

Opening an issue here after a little discussion in the discord server. 

Axum has an extractor for fetching connection information https://docs.rs/axum/latest/axum/struct.Router.html#method.into_make_service_with_connect_info but there isn't a way currently to get at it.

Having access to the client IP address unlocks many use cases including localization, regional age verification, etc.

### Describe the solution you'd like

Either having that information included with the header and body by default on the `Request` struct, or a way to manually add extractors to the handler methods. 

### Additional context

I'm happy to help implement something!

PR #754, opened 2025-04-23
feat: allow for custom server logic

aig787:
Replaced `TcpListener` with a generic `serve` function to allow more flexibility in server initialization. Updated `App` and template logic to detect and configure server functionality dynamically. This change enhances modularity and simplifies future extensions.

### Checklist

- [X] I have read [Contributing > Pull requests](https://tuono.dev/documentation/contributing/pull-requests)

### Related issue

Partially addresses #753 

### Overview

This change adds the ability to specify custom serve logic. I followed the pattern of custom app state, and the current behavior is preserved if a `src/server.rs` file is not found. The contract of the custom server function is:
```rust
async fn main(server_address: String, router: Router) -> Result<(), std::io::Error> {
    // custom logic here
}
```

I went back and forth on whether it was better to pass in a `String` or `TcpListener` here, but ultimately did it this way because in the past I've run across libraries where there wasn't any way to customize the listener (e.g. to use rustls over tls) and that's been a hinderance. 

Is this on the right track? Any additional tests or documentation I need to add? I did break the mock server, but I wanted to validate the approach before I fixed things downstream.

### DIFF
--- a/crates/tuono/src/app.rs
+++ b/crates/tuono/src/app.rs
@@ -40,6 +40,7 @@ pub struct App {
     pub route_map: HashMap<String, Route>,
     pub base_path: PathBuf,
     pub has_app_state: bool,
+    pub has_server: bool,
     pub config: Option<Config>,
 }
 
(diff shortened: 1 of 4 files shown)


Valerioageno:
What about giving to the final users the maximum flexibility here?
Instead of creating a `src/server.rs` just create a `src/main.rs` that needs to be entirely initialized using axum.

The only core functionalities to adapt are:
- initialize `v8` engine
- Load the file system generated route

I think this way is pretty cool since it solves all future needs. WDYT?

PR #755, opened 2025-04-24
feat: Add support passing in custom axum extractors

aig787:
Introduced `axum_arguments` attribute to the `handler` macro for selective inclusion of arguments. Added parsing, validation, and code generation logic to process these attributes.

### Checklist

### Related issue

Partially addresses #753 

### Overview

Added a new `axum_arguments` parameter to the `handler` macro that allows for passing in custom/other extractors. The arguments that should be treated as extractors are specified as:

```rust
#[tuono_lib::handler(axum_arguments(foo, bar))]
```

### DIFF
--- a/crates/tuono_lib_macros/Cargo.toml
+++ b/crates/tuono_lib_macros/Cargo.toml
@@ -4,14 +4,14 @@ version = "0.19.4"
 edition = "2024"
 description = "Superfast React fullstack framework"
 homepage = "https://tuono.dev"
-keywords = [ "react", "typescript", "fullstack", "web", "ssr"]
+keywords = ["react", "typescript", "fullstack", "web", "ssr"]
 repository = "https://github.com/tuono-labs/tuono"
 readme = "../../README.md"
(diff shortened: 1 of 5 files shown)


aig787:
Just pushed an update that changes the signature contract to 

```
#[tuono_lib::handler(axum_arguments(
    (cookies),
    (query, extractor = Query)
))]
async fn get_hello(req: Request, query: MyQuery, cookies: CookieJar) -> Response {
```

The previous version blew up if you passed in `Query(query): Query<MyQuery>` as an argument, and since it's an attribute macro, I can't replace that with the simplified `query: MyQuery` representation. This lets you pass those details on the macro instead of in the function signature.

Valerioageno:
I feel the syntax here is a bit awkward.
Do you think it would be possible managing `render_to_string()` and `json()` methods within a [tower service](https://docs.rs/tower/latest/tower/)?  
This way, we would just have to wrap the user defined handlers within the service and leaving to the end user the full axum handlers supports. WDYT

PR #764, opened 2025-05-05
feat: Allow axum extractor in tuono handlers

### LINKED ISSUE #761: [Feature request]: Allow axum extractor in tuono handlers

Refer to #526

Refer to #526

Remove `tuono_lib::Request` and allow every user defined handler to take as argument any axum extractor.

> Note: the handlers should still return `tuono_lib::Response`

_No response_

### Additional context

_No response_

aig787:
### Checklist

- [x ] I have read [Contributing > Pull requests](https://tuono.dev/documentation/contributing/pull-requests)

### Related issue

Fixes #761

### Overview

Removes `tuono_lib::Request` in favor of using `axum::extract::Request` directly. This allows handler methods to specify arbitrary extractors as parameters.


crates/tuono_lib_macros/src/handler.rs
```diff
@@ -1,77 +1,94 @@
...
-        }
+    let mut request_arg_index: Option<usize> = None;
 
+    for (i, arg) in item.sig.inputs.iter().enumerate() {
         if let FnArg::Typed(pat_type) = arg {
-            let index = i - 1;
-            let argument_name = *pat_type.pat.clone();
-            argument_names.insert(index, argument_name.clone());
+            match &*pat_type.pat {
+                Pat::Ident(PatIdent { ident, .. }) => {
```
aig787, line 21: This section could use a second pair of eyes. Here I check each of the parameters to see if it's a request argument and save it off. I check the last segment so if the user specified `axum::extract::Request` it would still work _but_ if they specified `some::other::Request` it would take that as the request argument. I'm not sure of a good way around that or if it matters.

We need to know if there's a request argument and what it's called so that we can pass it into `render_to_string()`. I did look into making this a tower service, but the mechanics of how Handler interacts with a Tower layer needed more time to wrap my head around, and this approach doesn't have the runtime overhead of a layer.

aig787: Ok, pipelines should be fixed now the e2e tests pass locally for me

crates/tuono_lib/Cargo.toml
```diff
@@ -5,35 +5,35 @@ edition = "2024"
...
-		"Cargo.toml"
+    "Cargo.toml"
 ]
 
 
 [dependencies]
 ssr_rs = "0.8.2"
-axum = {version = "0.8.1", features = ["json", "ws"]}
-axum-extra = {version = "0.10.0", features = ["cookie"]}
+axum = { version = "0.8.1", features = ["json", "ws", "form"] }
```
Valerioageno, line 21: I'd import all the FF

crates/tuono_lib/tests/utils/form_data.rs
```diff
@@ -1,13 +1,12 @@
+use axum::extract::Form;
 use serde::Deserialize;
-use tuono_lib::Request;
 
 #[derive(Deserialize)]
-struct Payload {
+pub struct Payload {
```
Valerioageno, line 5: does it need to be `pub`?

crates/tuono_lib/tests/utils/post_api.rs
```diff
@@ -1,12 +1,12 @@
+use axum::extract::Json;
 use serde::Deserialize;
-use tuono_lib::Request;
 
 #[derive(Deserialize)]
-struct Payload {
+pub struct Payload {
```
Valerioageno, line 5: Same here

crates/tuono_lib_macros/src/api.rs
```diff
@@ -22,66 +19,34 @@ pub fn api_core(attrs: TokenStream, item: TokenStream) -> TokenStream {
...
         #item
 
-        pub async fn #api_fn_name(#axum_arguments)#return_type {
-
-           #application_state_extractor
-
-           #modified_request
-
-           #fn_name(req.clone(), #argument_names).await
+        pub async fn #api_fn_name(#arguments)#return_type {
```
Valerioageno, line 48: Couldn't we just rename the function?

crates/tuono_lib_macros/src/api.rs
```diff
@@ -22,66 +19,34 @@ pub fn api_core(attrs: TokenStream, item: TokenStream) -> TokenStream {
...
-
-        if i == 1 {
-            axum_arguments.insert(1, create_struct_fn_arg())
-        }
-
         if let FnArg::Typed(pat_type) = arg {
-            let index = i - 1;
-            let argument_name = *pat_type.pat.clone();
-            argument_names.insert(index, argument_name.clone());
+            match &*pat_type.pat {
```
Valerioageno, line 27: Given the complexity of the approach, I think a good situation for adding tests

crates/tuono_lib_macros/src/api.rs
aig787: I'm not sure I follow

crates/tuono_lib/tests/utils/form_data.rs
aig787: They could be `pub(crate)`, but since they're test files I didn't worry about it. They need to be visible to the mock server or you get 

```
error: type `form_data::Payload` is private
  --> crates/tuono_lib/tests/utils/mock_server.rs:90:43
   |
90 |             .route("/api/form_data", post(form_data_api))
   |                                           ^^^^^^^^^^^^^ private type
```

PR #780, opened 2025-05-18
fix(crates/tuono): Fix HTTP method handling for mixed-case input

aig787:
Ensure HTTP methods are processed in uppercase to prevent mismatches. Added tests to validate the handling of routes with mixed-case HTTP method annotations and confirmed correct generation of imports.

### Checklist

### Related issue

None

### Overview

Discovered kind of a fun bug testing my other changes. You'd think a  `HashSet<Method>` would only store unique http methods, but for reasons I don't really understand - that isn't honored when you use `Method::from_str()`. The effect of this in tuono is that if you had macros defined for api with mixed case, e.g. `#[handler(GET)]` and `#[handler(get)]` it would work fine, except you get duplicate imports in the generated main. This case would hit anyone who used the lowercase notation in a bootstrapped project because the templated healthcheck api uses uppercase.

### DIFF
--- a/crates/tuono/src/app.rs
+++ b/crates/tuono/src/app.rs
@@ -253,7 +253,6 @@ impl App {
 
 #[cfg(test)]
 mod tests {
-
     use super::*;
 
     #[test]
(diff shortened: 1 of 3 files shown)


Valerioageno (approved):
Thanks for the improvement :)

